PR #18, opened 2016-10-06
Update to work with latest analyzer.

vikerman:
Use AnalysisDomain.onResultChanged instead of onResultComputed.
Update dependency on package plugin to match that of the latest analyzer.


server_plugin/lib/src/analysis.dart
```diff
@@ -16,7 +16,7 @@ import 'package:angular2_analyzer_plugin/src/tasks.dart';
 class AnalysisDomainContributor {
   AnalysisDomain analysisDomain;
 
-  void onResult(ComputedResult result) {
+  void onResult(ResultChangedEvent result) {
```
scheglov: Rename "result" to "event".

And wrap the whole body into `if (event.wasComputed) {}`.

PR #20, opened 2016-10-12
Fix an exception when processing local variables in templates.

vikerman:
The newly created local variable needs to be added to the localVariables list of the enclosing element or it would cause an exception later while getting its hash code(for putting in a map) later.

Fixes 12 of the 14 failing tests.


scheglov:
LGTM

PR #21, opened 2016-10-19
Fix check for Angular 2 annotations.

vikerman:
Check whether an annotaion belongs to Angular 2 using the library source path. The library names for Angular 2 sources are empty because there are no library declarations in Angular 2.

### DIFF
--- a/analyzer_plugin/lib/src/tasks.dart
+++ b/analyzer_plugin/lib/src/tasks.dart
@@ -1126,7 +1126,7 @@ class _AnnotationProcessorMixin {
   bool _isAngularAnnotation(ast.Annotation node, String name) {
     if (node.element is ConstructorElement) {
       ClassElement clazz = node.element.enclosingElement;
-      return clazz.library.name == 'angular2.src.core.metadata' &&
+      return clazz.library.source.uri.path.contains('angular2') &&
           clazz.name == name;
     }
     return false;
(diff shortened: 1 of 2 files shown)


vikerman:
Please review: @scheglov @ddrone

analyzer_plugin/lib/src/tasks.dart
```diff
@@ -1126,7 +1126,7 @@ class _AnnotationProcessorMixin {
   bool _isAngularAnnotation(ast.Annotation node, String name) {
     if (node.element is ConstructorElement) {
       ClassElement clazz = node.element.enclosingElement;
-      return clazz.library.name == 'angular2.src.core.metadata' &&
+      return clazz.library.source.uri.path.contains('angular2') &&
```
ddrone: IMO it would be better to check whether fqcn starts with "angular2." as this check might succeed for unintended libraries (say, "foocompany.helpers.angular2")

scheglov: I agree, the check is too wide now.
Maybe check that the path ends with `angular2/lib/src/core/metadata.dart`.

vikerman: Done. PTAL

vikerman:
Made changes. PTAL

PR #23, opened 2016-10-28
Support @Components without @Views as the default case, require template

MichaelRFairhurst:
In the view building task, check for @Component without a @View and
handle it. Also make sure that the @View or @Component define a template
or templateUrl but not both. Tests included.

Changed most tests that used @View + @Component to just use @Component
since that is the primary use case. Kept two positive test cases for
@View + @Component, adding corresponding ones for simply @Component by
itself. Hopefully those are adequate for ensuring support for @View,
since the code is (at least currently) very similar for both workflows.


MichaelRFairhurst:
I signed it!

Well, really I just set my google email address on my github and I'm hoping it figures out I'm an employee from that.

zoechi:
I wasn't aware `@View()` still exists in Dart. What's the purpose?

PR #28, opened 2016-10-31
Detect missing html template files in components

MichaelRFairhurst:
Plus tests.

Previous approach was overly complicated, thanks to Konstantin Shcheglov
for suggesting this much much simpler one.

A prior test with an invalid html template was now throwing this new error in addition to what it expected. In the name of "exactly one test should fail" I set that test to have a valid file.

### DIFF
--- a/analyzer_plugin/test/tasks_test.dart
+++ b/analyzer_plugin/test/tasks_test.dart
@@ -630,10 +630,11 @@ class ComponentA {
         r'''
 import '/angular2/angular2.dart';
 
-@Component(selector: 'aaa', template: 'AAA', templateUrl: 'AAA')
+@Component(selector: 'aaa', template: 'AAA', templateUrl: 'a.html')
 class ComponentA {
 }
 ''');
+    newSource('/a.html', '');
     LibrarySpecificUnit target = new LibrarySpecificUnit(source, source);
     computeResult(target, VIEWS);
     expect(task, new isInstanceOf<BuildUnitViewsTask>());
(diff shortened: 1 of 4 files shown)


analyzer_plugin/test/tasks_test.dart
```diff
@@ -662,6 +663,29 @@ class ComponentA {
         <ErrorCode>[AngularWarningCode.NO_TEMPLATE_URL_OR_TEMPLATE_DEFINED]);
   }
 
+  void test_hasError_missingHtmlFile() {
+    String code = r'''
+import '/angular2/angular2.dart';
+
+@Component(selector: 'my-component', templateUrl: 'my-template.html')
```
scheglov: I usually use the value that says itself that it does not exist :-)
`templateUrl: 'does-not-exist.html'` FYI, you don't have to do this change :-)

analyzer_plugin/test/tasks_test.dart
```diff
@@ -662,6 +663,29 @@ class ComponentA {
...
+    computeResult(target, VIEWS);
+    expect(task, new isInstanceOf<BuildUnitViewsTask>());
+    // validate
+    fillErrorListener(VIEWS_ERRORS);
+    errorListener.assertErrorsWithCodes(<ErrorCode>[
+            AngularWarningCode.REFERENCED_HTML_FILE_DOESNT_EXIST]);
+    {
+      AnalysisError error = errorListener.errors.single;
+      expect(error.offset, code.indexOf("'my-template.html'"));
+      expect(errorListener.errors.single.length, "'my-template.html'".length);
```
scheglov: Checking for position make the test more complete, that's nice.
If we want to do this often, it might be worth to have a utility function.

MichaelRFairhurst:
Should I be able to merge this now?

scheglov:
Yes, of course!

MichaelRFairhurst:
Hmm, I only see "comment" and "close and comment"

matanlurey:
@MichaelRFairhurst We might need to give you push privileges. /cc @kevmoo

PR #34, opened 2016-11-10
Template errors beyond first tag

MichaelRFairhurst:
Noticed that the error wasn't occurring inside full html templates, saw
the difference in how the dart analyzer handles html parsing and applied
it here. Saves us from getting highly involved in the dart html parser
API when the root isn't simple.


MichaelRFairhurst:
Not sure why this travis job is still pending...

PR #63, opened 2016-11-17
Fix issue of template attrs being resolved twice, once without directives

MichaelRFairhurst:
This only applies to `<template ngFor...>`, note that the sugared template
syntaxes are handled differently and don't have this problem because the
sugared attributes are handled before the other attributes are.

No longer resolve template attributes before the template directives are
resolved. Still, we have to resolve the ngFor attribute before we can
put the ngFor variables in scope. So, split that apart from the other
types of vars that are defined based on directives. In the case of
resolving attributes for a template, do the ngFor resolution later but
not too late; in the case of sugared templates, do the same order.

Edit: fixes two tests: https://travis-ci.org/dart-lang/angular_analyzer_plugin/jobs/176588080 failure caused be the addition of input validation, which were revealed after merged with ngFor support since it added assertNewErrors to those tests

### DIFF
--- a/analyzer_plugin/lib/src/resolver.dart
+++ b/analyzer_plugin/lib/src/resolver.dart
@@ -478,6 +478,10 @@ class TemplateResolver {
       internalVariables[r'$implicit'] = new InternalVariable(
           r'$implicit', new DartElement(classElement), classElement.type);
     }
+  }
+
+  void _defineNgForVariables(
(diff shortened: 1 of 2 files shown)


scheglov:
Applies to... what?

MichaelRFairhurst:
Ah, yes, I didn't specify.

Fixes two tests: https://travis-ci.org/dart-lang/angular_analyzer_plugin/jobs/176588080 :)

Issue #41, opened 2016-11-10
Ensure outputs are eventemitters


MichaelRFairhurst:
applies to #38  and #39 but I think warrants its own ticket

MichaelRFairhurst:
PR #68 merged

Issue #37, opened 2016-11-10
Check that $event is used correctly (placement + types)


MichaelRFairhurst:
Currently all dart expressions accept $event as an 'any' type, but we can do better than that.

MichaelRFairhurst:
PR #68 merged

PR #68, opened 2016-11-18
Typecheck $event (#37) and ensure @Output are EventEmitters (#41)

MichaelRFairhurst:
The parseDartExpression code defines a local scope, so have it accept a
nullable EventType thing to fill in the $event var on that scope rather
than have stuff outside of that function have to set up/tear down the
scope properly.

Track the event types of html components as well as the event types of
custom components, and store those in OuputElement since they have
different methods of being discovered

### DIFF
--- a/analyzer_plugin/lib/src/model.dart
+++ b/analyzer_plugin/lib/src/model.dart
@@ -1,6 +1,7 @@
 library angular2.src.analysis.analyzer_plugin.src.model;
 
 import 'package:analyzer/dart/element/element.dart' as dart;
+import 'package:analyzer/dart/element/type.dart' as dart;
 import 'package:analyzer/src/generated/source.dart' show Source, SourceRange;
 import 'package:analyzer/src/generated/utilities_general.dart';
 import 'package:analyzer/task/model.dart' show AnalysisTarget;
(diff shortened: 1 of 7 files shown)


analyzer_plugin/lib/src/model.dart
```diff
@@ -204,7 +207,7 @@ class OutputElement extends AngularElementImpl {
   final SourceRange getterRange;
 
   OutputElement(String name, int nameOffset, int nameLength, Source source,
-      this.getter, this.getterRange)
+      this.getter, this.getterRange, this.eventType)
```
scheglov: Last time when I thought about events, it seemed to me that any `Stream<SomeEvent>` can be used.

MichaelRFairhurst: Hmm, allowing supertypes in general is important...especially if any stream is allowable. We have a ticket for that but it might be important to up its priority

scheglov: I don't know about priorities.
Just saying that it might be wider.
It's perfectly fine IMHO to have a less general, but reasonable implementation.

MichaelRFairhurst: Ah yes. Definitely important that we know exactly how wide it is or isn't!

Issue #50, opened 2016-11-11
two way binding validation


MichaelRFairhurst:
For instance, it would be nice to find ([thisError]), and we can do the usual type checking

MichaelRFairhurst:
PR #79  merged!

PR #79, opened 2016-11-21
Issue 50 twoway binding validation

MichaelRFairhurst:
Support for #50 two-way binding validation

validate types (as an input), assignabliity, and change the way unbound
error types are tracked. For a two-way bound input named x of type t,
check that an ouput name xChange exists which is assignable to type t.

Smarter tactic of ignoring :not()s, so that core directives work

### DIFF
--- a/analyzer_plugin/lib/src/resolver.dart
+++ b/analyzer_plugin/lib/src/resolver.dart
@@ -35,7 +35,7 @@ html.Element _firstElement(html.Node node) {
   return null;
 }
 
-enum AttributeBoundType { input, output }
+enum AttributeBoundType { input, output, twoway }
 
 /**
  * Information about an attribute.
@@ -300,7 +300,11 @@ class HtmlTreeConverter {
         AttributeBoundType bound = null;
         String propName = name;
         int propNameOffset = nameOffset;
-        if (propName.startsWith('[') && propName.endsWith(']')) {
+        if (propName.startsWith('[(') && propName.endsWith(')]')) {
+          propNameOffset += 2;
+          bound = AttributeBoundType.twoway;
+          propName = propName.substring(2, propName.length - 2);
+        } else if (propName.startsWith('[') && propName.endsWith(']')) {
           propNameOffset += 1;
           propName = propName.substring(1, propName.length - 1);
           bound = AttributeBoundType.input;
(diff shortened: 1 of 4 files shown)


analyzer_plugin/lib/src/resolver.dart
```diff
@@ -35,7 +35,7 @@ html.Element _firstElement(html.Node node) {
   return null;
 }
 
-enum AttributeBoundType { input, output }
+enum AttributeBoundType { input, output, twoway }
```
scheglov: Is `twoway` a word?
Might be better to use `twoWay`.

analyzer_plugin/lib/tasks.dart
```diff
@@ -116,7 +116,27 @@ class AngularWarningCode extends ErrorCode {
...
 
   /**
-   * An error code indicating that a nonexist input was bound
+   * An error code indicating that a nonexist output was bound, perhaps
+   * because an input was two way bound. The nonexist bound output is
+   * an implementation detail, so give its own error.
+   */
+  static const AngularWarningCode NONEXIST_TWOWAY_OUTPUT_BOUND =
+      const AngularWarningCode('NONEXIST_TWOWAY_OUTPUT_BOUND',
+          'The two way binding {0} requires a bindable output of name {1}');
```
scheglov: `two-way` is used below.

analyzer_plugin/test/resolver_test.dart
```diff
@@ -165,6 +165,136 @@ class TestPanel {
         AngularWarningCode.NONEXIST_INPUT_BOUND, code, "[title]");
   }
 
+  void test_expression_twowayBinding_valid() {
```
scheglov: If you decide to change to `twoWay`, you will need to update also test method names.

analyzer_plugin/test/resolver_test.dart
```diff
@@ -165,6 +165,136 @@ class TestPanel {
...
+@Directive(selector: '[titled]', template: '', inputs: 'title')
+class TitleComponent {
+  @Input() String title;
+  @Output() EventEmitter<String> titleChange;
+}
+''');
+    _addHtmlSource(r"""
+<span titled [(title)]='text'></span>
+""");
+    _resolveSingleTemplate(dartSource);
```
scheglov: It might be also interesting to verify that `title` in `[(title)]` is resolved to... not sure... input?

Input and output should be bound together and when rename refactoring renames one, it should also rename the other. Very long term goal though :-)

analyzer_plugin/lib/src/resolver.dart
MichaelRFairhurst: Probably, yes, twoWay is better. It's not quite "two way", its really "two-way"...but most style guides make "xY" the brother of "x-y", not "xy." So while "two-way" is not a css rule, its still a precedent that I'm happy to follow. Change incoming.

Issue #77, opened 2016-11-20
"keyup" event on input tag not working


MichaelRFairhurst:
I think its because the dart:html event doesn't begin with "on" and we are stripping it, but it could be some other reason.

```<input (keyup)="code" />```

getting error "the bound output does not exist on any directives"

**Comment by MichaelRFairhurst** [comment]
Also seems to be an issue with mouseenter and mouseleave

**Comment by zoechi** [comment]
> the bound output does not exist on any directives"

I think it just shouldn't expect DOM events being exposed as outputs

**Comment by MichaelRFairhurst** [comment]
It doesn't yet get to the point of recognizing this as a DOM event. Anything that isn't a DOM event it expects to be an output, and so that's why it complains about the output errors.

Turns out in dart:html, the event is described like so:

```
    @DomAttribute("onkeyup")
    ElementStream<KeyEvent> get onKeyUp() {
```

We analyze that, but a subtly wrong way. We take the method name and lop off the "on" and decapitalize it to get "keyUp." Therefore

```<input (keyUp)="code" />```

works just fine.

The solution is to analyze the `@DomAttribute` annotation instead of the method name.

MichaelRFairhurst:
Merge PR #80

PR #80, opened 2016-11-22
Fix for issue #77 some dom events not recognized

MichaelRFairhurst:
Don't take the dart:html method name and chop off the 'on', that's not
how angular does it. Instead take the @DomName('Document.oneventname')
and get just the eventname out of that.

Previous tests used 'keyUp' as a testcase, but the correct angular is
'keyup'. Note that this difference is in dart:html and was kept in
mock_sdk.dart

### DIFF
--- a/analyzer_plugin/lib/src/tasks.dart
+++ b/analyzer_plugin/lib/src/tasks.dart
@@ -1382,7 +1382,8 @@ class _BuildStandardHtmlComponentsVisitor extends RecursiveAstVisitor {
 
   String _getDomName(Element element) {
     for (ElementAnnotation annotation in element.metadata) {
-      var value = annotation.constantValue ?? annotation.computeConstantValue();
+      // this has caching built in, so we can compute every time
+      var value = annotation.computeConstantValue();
       if (value != null && value.type is InterfaceType) {
         if (value.type.element.name == 'DomName') {
           return value.getField("name").toStringValue();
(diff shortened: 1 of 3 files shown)


analyzer_plugin/lib/src/tasks.dart
```diff
@@ -1371,6 +1380,19 @@ class _BuildStandardHtmlComponentsVisitor extends RecursiveAstVisitor {
     });
   }
 
+  String _getDomName(Element element) {
+    for (ElementAnnotation annotation in element.metadata) {
+      var value = annotation.constantValue ?? annotation.computeConstantValue();
```
scheglov: I think `computeConstantValue` already does caching internally.

```dart
  /**
   * Return a representation of the value of this variable, forcing the value
   * to be computed if it had not previously been computed, or `null` if either
   * this variable was not declared with the 'const' modifier or if the value of
   * this variable could not be computed because of errors.
   */
  DartObject computeConstantValue() {
    if (evaluationResult == null) {
      context?.computeResult(this, CONSTANT_VALUE);
    }
    return evaluationResult?.value;
  }
```

analyzer_plugin/test/tasks_test.dart
```diff
@@ -57,6 +58,19 @@ class BuildStandardHtmlComponentsTaskTest extends AbstractAngularTest {
...
         expect(input.setter, isNotNull);
       }
+      {
+        InputElement input = inputs.singleWhere((i) => i.name == 'tabIndex');
+        expect(input, isNotNull);
+        expect(input.setter, isNotNull);
+      }
+      {
+        OutputElement output =
+            outputElems.singleWhere((o) => o.name == 'click');
```
scheglov: Your attempt to make `outputElements` shorter did not help, the line is still wrapped.
Maybe it does not worth it :-)

Issue #76, opened 2016-11-20
Allow subtypes of EventEmitters in @Outputs


MichaelRFairhurst:
Currently, code such as

```
class MyEventEmitter<T> extends EventEmitter<T> { ... }

...
   @Output MyEventEmitter<int> intEvents;
```

will fail.

Per @scheglov, its possible that `EventEmitter` itself is already a subtype of the required type, which may be simply `Stream<T>`.

Note as well that we should do this right, tracking where the subtype relationship specifies the type parameters, ie

```class MyIntEventEmitter extends EventEmitter<int> { ... }```

If this is considerably more work then a new ticket should be made for this edge case.

MichaelRFairhurst:
Upping the importance of this for two reasons: Firstly, per some discovery showing it comes up a lot. Secondly, when this goes wrong it currently has no fallback behavior and it complains about the $event variables when it can't identify their type.

Also confirmed that `Stream` is the required type, not `EventEmitter`.

Therefore the goals are twofold:

1. support subtypes of `Stream<T>` while allowing for classes to specialized Streams (ie, `extends Stream<int>`), and
2. use `dynamic` type when the event type can't be found

scheglov:
For `Stream` subclasses you might want to look at https://github.com/dart-lang/sdk/blob/1ac36f7c3872911b897fc1c61ac90bd9489d650d/pkg/analyzer/lib/src/generated/type_system.dart#L1193 and its use in https://github.com/dart-lang/sdk/blob/3c437f5d9e3b2cefd581d34f08d8ab4484bea608/pkg/analyzer/lib/src/generated/error_verifier.dart#L4220.

MichaelRFairhurst:
Ah! That is exactly what I was hoping would exist!

MichaelRFairhurst:
PR #89 merged

PR #89, opened 2016-11-28
Allow Stream<T> and subtypes of it in outputs, refs #76

MichaelRFairhurst:
Use the mostSpecificTypeArgument method API to get the stream type
regardless of the inheritance tree and generics defined within it.

### DIFF
--- a/analyzer_plugin/lib/src/tasks.dart
+++ b/analyzer_plugin/lib/src/tasks.dart
@@ -228,6 +228,7 @@ class BuildStandardHtmlComponentsTask extends AnalysisTask {
 class BuildUnitDirectivesTask extends SourceBasedAnalysisTask
     with _AnnotationProcessorMixin {
   static const String UNIT_INPUT = 'UNIT_INPUT';
+  static const String TYPE_PROVIDER_INPUT = 'TYPE_PROVIDER_INPUT';
 
   static final TaskDescriptor DESCRIPTOR = new TaskDescriptor(
       'BuildUnitDirectivesTask',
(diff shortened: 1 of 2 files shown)


analyzer_plugin/lib/src/tasks.dart
```diff
@@ -437,9 +441,11 @@ class BuildUnitDirectivesTask extends SourceBasedAnalysisTask
     if (getter != null && getter.type != null) {
       var returntype = getter.type.returnType;
```
scheglov: `returnType`?

analyzer_plugin/test/tasks_test.dart
```diff
@@ -620,6 +620,124 @@ class MyComponent {
     errorListener.assertNoErrors();
   }
 
+  void test_outputs_streamIsOk() {
+    String code = r'''
+import '/angular2/angular2.dart';
+import 'dart:async';
+
+@Component(
+    selector: 'my-component',
```
scheglov: I think this is a compilation error.
Here and below.

analyzer_plugin/lib/src/tasks.dart
MichaelRFairhurst: Great catch. Looks like that occurred lower in the file too, fixed both

analyzer_plugin/test/tasks_test.dart
MichaelRFairhurst: Wow, lots of these. Was just doing a merge and saw this. Thought I just read the diff wrong! fixed

analyzer_plugin/lib/src/tasks.dart
```diff
@@ -1390,14 +1390,14 @@ class _BuildStandardHtmlComponentsVisitor extends RecursiveAstVisitor {
 
       if (!outputMap.containsKey(name)) {
         if (accessor.isGetter) {
-          var returntype =
+          var returnType =
               accessor.type == null ? null : accessor.type.returnType;
           DartType eventtype = null;
```
scheglov: `eventType`

PR #90, opened 2016-11-28
Issue 36 pull request

mk13:
Fix now properly checks for for end of expression via 'eof' token and marks any subsequent string as erroneous.

### DIFF
--- a/analyzer_plugin/lib/src/resolver.dart
+++ b/analyzer_plugin/lib/src/resolver.dart
@@ -1061,12 +1061,18 @@ class TemplateResolver {
         textOffset = begin + 2;
         continue;
       }
-
       // resolve
       begin += 2;
       String code = text.substring(begin, end);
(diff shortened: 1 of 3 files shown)


analyzer_plugin/lib/src/resolver.dart
```diff
@@ -1212,12 +1212,18 @@ class TemplateResolver {
...
       }
-
       // resolve
       begin += 2;
       String code = text.substring(begin, end);
-      _resolveExpression(fileOffset + begin, code);
+      Expression expression = _resolveExpression(fileOffset + begin, code);
       textOffset = end + 2;
+
+      if (expression.endToken.next.type != TokenType.EOF){
```
scheglov: 1. Extract `expression.endToken.next`.
2. We already know the span of the code - from `begin` to `end`, don't we? So, the rest after `expression` is from `expression.end` to `end`.

PR #94, opened 2016-11-28
Issue 36 resolve issue where errors aren't detected for trailing expr…

mk13:
Issue 46; re-submit commits and pull request under proper author

### DIFF
--- a/analyzer_plugin/lib/src/resolver.dart
+++ b/analyzer_plugin/lib/src/resolver.dart
@@ -986,10 +986,11 @@ class TemplateResolver {
     _recordExpressionResolvedRanges(expression);
 
     if (expression.endToken.next.type != TokenType.EOF) {
+      int trailingExpressionBegin = expression.endToken.next.offset;
       errorListener.onError(new AnalysisError(
           templateSource,
-          expression.endToken.next.offset,
-          offset + code.length - expression.endToken.next.offset,
+          trailingExpressionBegin,
+          offset + code.length - trailingExpressionBegin,
           AngularWarningCode.TRAILING_EXPRESSION));
     }
     return expression;
--- a/analyzer_plugin/lib/src/resolver.dart
+++ b/analyzer_plugin/lib/src/resolver.dart
@@ -1212,12 +1212,19 @@ class TemplateResolver {
         textOffset = begin + 2;
         continue;
       }
-
       // resolve
       begin += 2;
       String code = text.substring(begin, end);
-      _resolveExpression(fileOffset + begin, code);
+      Expression expression = _resolveExpression(fileOffset + begin, code);
       textOffset = end + 2;
+
+      if (expression.endToken.next.type != TokenType.EOF) {
+        errorListener.onError(new AnalysisError(
+            templateSource,
+            expression.endToken.next.offset,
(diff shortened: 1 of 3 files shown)


analyzer_plugin/lib/src/resolver.dart
```diff
@@ -1212,12 +1212,19 @@ class TemplateResolver {
...
       String code = text.substring(begin, end);
-      _resolveExpression(fileOffset + begin, code);
+      Expression expression = _resolveExpression(fileOffset + begin, code);
       textOffset = end + 2;
+
+      if (expression.endToken.next.type != TokenType.EOF) {
+        errorListener.onError(new AnalysisError(
+            templateSource,
+            expression.endToken.next.offset,
+            _computeTokenLengthToEOF(expression.endToken.next),
```
scheglov: Can we use `expression.end` to `end` instead?

MichaelRFairhurst: We're working on this, and also probably moving the error into parseExpression instead of in handleTextExpressions

MichaelRFairhurst:
LGTM, though I don't seem to have permission to approve!

analyzer_plugin/test/resolver_test.dart
```diff
@@ -723,6 +756,7 @@ class GenericComponent<T extends String> {
 ''');
     var code = r"""
 <generic-comp [(string)]="anInt"></generic-comp>
+>>>>>>> 593afa40db6a1638b59d840716ca66bcc846f136
```
scheglov: I think you want to remove this.

analyzer_plugin/lib/src/resolver.dart
```diff
@@ -984,6 +984,14 @@ class TemplateResolver {
   Expression _resolveExpression(int offset, String code) {
     Expression expression = _resolveDartExpressionAt(offset, code, null);
     _recordExpressionResolvedRanges(expression);
+
+    if (expression.endToken.next.type != TokenType.EOF) {
+      errorListener.onError(new AnalysisError(
+          templateSource,
+          expression.endToken.next.offset,
```
scheglov: This expression is used twice.
Could you please extract it into a local variable?

MichaelRFairhurst:
Before we merge this, I would like to (if Max you can forgive me for this...) set one last requirement before this is merged!

Our top goal is to prevent false errors, and I know we had decided to push back the statements-in-events part of issue 36, and make it a separate issue.

I realized today that if we merge this, the events in `(output)="bindings"`are currently parsed via `parseExpression`, and will err out. Yes, I told you to move the code into parseExpression so this is 100% my fault!

Let's talk tomorrow about the best way to get this merged, without causing false errors on outputs that _do_ accept more than just expressions which triggers our new error. Even if it doesn't mean doing all the proper handling on output statements just yet!

mk13:
Sounds good - I'm analyzing it right now and let's discuss the best course of action.

mk13:
Analysis shows that in (1)event statements, parsing stops and doesn't catch valid errors after first expression. However, it correctly ignores "trailing expression", which are valid in event statements. (2) In input-binding, it currently does not catch trailing expressions - but it should be. 

Merging this pull request since it causes no false errors.

PR #105, opened 2016-11-30
Refactoring TRAILING_EXPRESSION error into a more logical position

mk13:
Previous modification (Issue 36) only impacted expressions within moustache. This change will also now impact property binding expressions. There is a quick flag to prevent false-positives on event binding expressions; this is temporary until event binding expressions are properly implemented.


MichaelRFairhurst:
LGTM, going off of whether `eventType` is null is definitely a bit hacky but will do exactly what we want until events have their own statements parser.

mk13:
Agreed - I tried to look for a more elegant solution, but the technology wasn't there yet! I just want to ensure that for dogfooding purposes, there aren't unnecessary flags that bother the testers.

Issue #101, opened 2016-11-29
Bug? (mouseover) on a component


MichaelRFairhurst:
Need info if this is a bug (in other words, step one is to play around with this in a live angular app).

Seeing a project where (mouseover) is bound on a component:

```
   <custom-list-component (mouseover)="action">
   ...
```

I have a feeling this _does_ work, and that it binds to the outermost tag of the component. Ie,

```
@Component(template: "<div>...</div>", ...)
class MyComponent {
```

would have all the dom events that `div` does. If so, this could probably recurse, if the first element of `MyComponent` is a different component which has a template that starts with an html tag or a component that..... etc.

That said, this _might_ just be a bug in the codebase we were analyzing, that leads to a noop.

matanlurey:
Native events get caught up the DOM tree, like normal DOM would.

i.e.

```html
<div (click)="doSomething()">
  <button>Click me, it will still trigger doSomething</button>
</div>
```

Only native events have this property, though.

MichaelRFairhurst:
Great!

Looks like the DOM API already accounts for this; all bubbling events are present on the single type `HtmlElement`, since any element can bubble events from their children. We should be able to just make every component inherit all the outputs of HtmlElement from `dart:html` from the analyzer's perspective to solve this.

MichaelRFairhurst:
Merged PR #106

PR #106, opened 2016-11-30
Track native DOM events on componets refs #101

MichaelRFairhurst:
New `ResultDescriptor` to track the outputs of `HtmlElement` separately,
all elements at resolution stage can use them. No longer tracked on the
standard HTML components (now only a few special html components, such
as body).

When resolving events, check the standard html event when there's no
match otherwise.

Altered Mock SDK to better reflect the real dart html source, namely
* HtmlElement is not the root of the inheritance tree
* BodyElement has events not on HtmlElement
* keyup event is on HtmlElement, not InputElement
* Element (parent of HtmlElement) has events

No. 1 and 4 are important for correct implementation of
`_captureAspects`. Number 3 was a mistake I'd made when adding outputs.
Number two is an edge case.

Store everything in a nicely indexed map, because that's probably what
we should be doing everywhere for performance.

Resolving source ranges in _resolveAttributeValues() is new, but that's
where I see the code going in the future anyway, since resolving the
values requires resolving the names, so the two steps should probably be
eventually merged.

### DIFF
--- a/analyzer_plugin/lib/src/tasks.dart
+++ b/analyzer_plugin/lib/src/tasks.dart
@@ -202,17 +202,17 @@ class BuildStandardHtmlComponentsTask extends AnalysisTask {
     // Build components in each unit.
     //
     Map<String, Component> components = <String, Component>{};
-    Map<String, OutputElement> standardEvents = <String, OutputElement>{};
+    Map<String, OutputElement> events = <String, OutputElement>{};
     for (ast.CompilationUnit unit in units) {
       Source source = unit.element.source;
-      unit.accept(new _BuildStandardHtmlComponentsVisitor(
-          components, standardEvents, source));
+      unit.accept(
+          new _BuildStandardHtmlComponentsVisitor(components, events, source));
     }
     //
     // Record outputs.
     //
     outputs[STANDARD_HTML_COMPONENTS] = components.values.toList();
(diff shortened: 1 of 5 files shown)


analyzer_plugin/lib/src/tasks.dart
```diff
@@ -189,14 +202,17 @@ class BuildStandardHtmlComponentsTask extends AnalysisTask {
     // Build components in each unit.
     //
     Map<String, Component> components = <String, Component>{};
+    Map<String, OutputElement> standardEvents = <String, OutputElement>{};
```
scheglov: I think `standard` is superfluous here.

analyzer_plugin/lib/src/tasks.dart
```diff
@@ -1357,16 +1381,24 @@ class _AnnotationProcessorMixin {
 
 class _BuildStandardHtmlComponentsVisitor extends RecursiveAstVisitor {
   final Map<String, Component> components;
+  final Map<String, OutputElement> standardEvents;
```
scheglov: ...and here.

analyzer_plugin/lib/src/tasks.dart
MichaelRFairhurst: good point!

Issue #103, opened 2016-11-30
Generics through inheritance are *still* unresolved


MichaelRFairhurst:
Maybe call this a low-P1. We have a codebase where this is the third most numerous false error, but at this point even the third most common error is only comes up so often.

```
class Generic<T> {
  T something;
}

@Component(...)
class Component extends Generic {
}
```

Using `Component` in a template yields errors

```
<comp [something]="5">...
```

Type int is not assignable to T.

T is an internal type to `Generic`, and has no business being used in the template typechecker. It should be resolved to bounds (in this case `dynamic` since there are no bounds) and inherited constraints before being typechecked.

MichaelRFairhurst:
Merged PR #111

Issue #91, opened 2016-11-28
Bad errors on generic bounding chains following fix for #84, strong mode incompatible


MichaelRFairhurst:
In the branch with fix for #84, the type checks are working properly for no strong mode, but the errors are not fully clear/resolved.

```
MyComponent<A extends B, B extends C, C extends String> {
  @Input A stringBound;
}
```

When you use this directive in a template, we don't have type parameters but we do have the lower bounds.

```
<my-component [stringBound]="'thisShouldWork'">...
```

With the fix for #86, pass in a string, without strong mode, you get no errors. So that works. Pass in an integer, you get an error, so that works. Even in strong mode, I believe outputs work when they work.

However, the errors you get will say "int is not assignable to B" because `stringBound` is of type A, and A extends B and that's as far as we resolved it. The typing works because we replaced "A" on the type which had no "bound" set, to B from the ClassElement which did, and that bound makes it more or less act like its own bound when not in strong mode.

However, when in strong mode, String is not assignable to T extends String, even though the reverse is true.

The problem comes when calling substitute2, and it doesn't substitute recursively. Not sure if the fix for this should be in substitute2 in the analyzer repo?

There are tests for this in `tasks_test.dart`, referencing this ticket number. Uncomment them to do the work.

scheglov:
@jmesserly @leafpetersen Who might know better about types in analyzer.

leafpetersen:
> However, when in strong mode, String is not assignable to T extends String, even though the reverse is true.

This is correct.  If `T` extends `String`, then `String` may not be a subtype of `T`.  Can you be more concrete about what code is generating the error, and what you expect to happen?

> The problem comes when calling substitute2, and it doesn't substitute recursively. Not sure if the fix for this should be in substitute2 in the analyzer repo?

I don't know what this means, can you give me a concrete example of what you expect to happen vs what is happening?

jmesserly:
Yeah I want to add on to Leaf's question ... could you link to the code where you're implementing this? Are you trying to see if the string value "thisShouldWork" can be assigned to the "stringBound" field?

MichaelRFairhurst:
The example I gave where 'A extends B, B extends C'

I'm accessing a member of the class, and getting its type. That type is `A`, which won't work in the template typechecking, and correctly so as you confirmed.

[...]

PR #111, opened 2016-11-30
Solve issue #103 generic inherited types still unresolved

MichaelRFairhurst:
Move to a class-instantiation paradigm; instantiate the type to bounds
before building the component (still using custom instantiation logic
while issue #91 is blocked), and use that to resolve getters/setters up
the inheritance tree.

No change to the @Input() and @Output() workflows because those don't
involve inheritance.

### DIFF
--- a/analyzer_plugin/lib/src/tasks.dart
+++ b/analyzer_plugin/lib/src/tasks.dart
@@ -260,10 +260,18 @@ class BuildUnitDirectivesTask extends SourceBasedAnalysisTask
 
   TypeProvider typeProvider;
 
-  // Since <my-comp></my-comp> represents an instantiation of MyComp,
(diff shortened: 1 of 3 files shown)


analyzer_plugin/lib/src/tasks.dart
```diff
@@ -297,6 +302,7 @@ class BuildUnitDirectivesTask extends SourceBasedAnalysisTask
   AbstractDirective _createDirective(
       ast.ClassDeclaration classDeclaration, ast.Annotation node) {
     ClassElement classElement = classDeclaration.element;
+    _instantiatedClassType = _instantiateClass(classElement);
```
scheglov: You could do the same with `ClassElement`.
For example `_currentElement` and `_currentType`.

Issue #36, opened 2016-11-10
Not all dart syntax errors caught in templates


MichaelRFairhurst:
Looks to me like we try to parse an expression, but don't check that the whole expression was consumed by the parser.

Examples:
```{{valid_dart; I can write anything here and it doesn't catch it}}```
```{{something...}}```

both of these are invalid but not caught

**Comment by MichaelRFairhurst** [comment]
note that event handlers should be parsing dart statements where many other things parse dart expressions

**Comment by MichaelRFairhurst** [comment]
I think all dart bindings are in one of these forms:

```
<div>{{formOne}}</div>
<div title="prefix-{{formTwo}}"></div>
<comp [input]="formThree"></comp>
<comp (output)="formFour"></comp>
<input [(ngModel)]="formFive"></comp>
<div [attr.x]="formSix"></comp>
```

Currently they are all parsed & resolved by the same code I believe. And every form but form 4 are all being handled correctly.

Form four is currently handled in `resolveAttributeValues` when `attribute.bound == AttributeBoundType.output`. It needs its own logic to support statements.

Example test cases:

```
<div (click)="this; is; valid"></div><!-- no errors with multiple statements -->
<div (click)="soundlyTypedCode; 1 + 'unsoundlyTypedCode`"><!-- not a syntax error, but a type error! -->
<div (click)="return x;"></div><!-- only ExpressionStatement should be allowed I think! -->
<div (click)="if(x) { };"></div><!-- only ExpressionStatement should be allowed I think! -->
<div (click)=";;;;;;;; x;"></div><!-- Extra semicolons should be allowed I think -->
<div (click)="x"></div><!-- Not having a semicolon should be allowed, I think-->
<div (click)="abc}"></div><!-- syntax error on '}' -->
```

We can pair program on this, I think it will use a lot of the same lessons that the other half taught us, but also use the parser in new ways and we'll probably have to validate what it returns us more than before

mk13:
Pull request #131 merged is included for this issue.

PR #112, opened 2016-12-01
Statement handlers refs #36

mk13:
Building on issue 36

Event handlers are able to take statements instead of expressions. I've created a similar stack for resolving statements and generalized _recordExpressionResolvedRanges and _resolveDartExpression to be more generic to resolve any AstNode (expression or statement). 

Added a separate path to resolve event statements within _resolveAttributeValues. This will be refactored in another pull request.

### DIFF
--- a/analyzer_plugin/lib/src/resolver.dart
+++ b/analyzer_plugin/lib/src/resolver.dart
@@ -715,6 +715,10 @@ class TemplateResolver {
         }
 
         Expression expression = attribute.expression;
+
+        //Check if bound == OUTPUT:
+        //  If so, branch off and deal as statement,
+        //  otherwise, continue stack
         if (expression == null) {
           expression = _resolveDartExpressionAt(valueOffset, value, eventType);
           attribute.expression = expression;
--- a/analyzer_plugin/lib/src/resolver.dart
+++ b/analyzer_plugin/lib/src/resolver.dart
@@ -644,9 +644,9 @@ class TemplateResolver {
   /**
    * Parse the given Dart [code] that starts ot [offset].
    */
-  List<Statement> _parseDartStatement(int offset, String code) {
+  List<Statement> _parseDartStatements(int offset, String code) {
     Token token = _scanDartCode(offset, code);
-    return _parseDartStatementAtToken(token);
+    return _parseDartStatementsAtToken(token);
   }
 
   /**
@@ -660,7 +660,7 @@ class TemplateResolver {
   /**
    * Parse the Dart statement starting at the given [token].
    */
-  List<Statement> _parseDartStatementAtToken(Token token) {
+  List<Statement> _parseDartStatementsAtToken(Token token) {
     Parser parser = new Parser(templateSource, errorListener);
     return parser.parseStatements(token);
   }
@@ -756,6 +756,23 @@ class TemplateResolver {
               template.addRange(range, standardHtmlEvent);
             }
           }
+
+          //TODO: Refactor the following chunk of statement resolver
+          List<Statement> statements = _resolveDartStatementsAt(valueOffset, value, eventType);
+          for (Statement statement in statements){
+            _recordAstNodeResolvedRanges(statement);
+          }
+
+          if (!matched && unboundErrorCode != null) {
+            errorListener.onError(new AnalysisError(
+                templateSource,
+                attribute.nameOffset,
+                attribute.name.length,
+                unboundErrorCode,
+                [attribute.propertyName]));
+          }
+
+          continue;
         }
 
         Expression expression = attribute.expression;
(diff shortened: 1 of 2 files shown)


analyzer_plugin/lib/src/resolver.dart
```diff
@@ -731,15 +747,37 @@ class TemplateResolver {
...
+          }
+
+          continue;
         }
 
         Expression expression = attribute.expression;
+
+        //Check if bound == OUTPUT:
+        //  If so, branch off and deal as statement,
+        //  otherwise, continue stack
```
MichaelRFairhurst: I think this is a leftover comment

analyzer_plugin/lib/src/resolver.dart
```diff
@@ -731,15 +747,37 @@ class TemplateResolver {
...
         if (expression == null) {
           expression = _resolveDartExpressionAt(valueOffset, value, eventType);
           attribute.expression = expression;
         }
         if (expression != null) {
-          _recordExpressionResolvedRanges(expression);
+          _recordAstNodeResolvedRanges(expression);
```
MichaelRFairhurst: Is _recordExpressionResolvedRanges still in the file, and is it still used anywhere?

analyzer_plugin/lib/src/resolver.dart
```diff
@@ -731,15 +747,37 @@ class TemplateResolver {
               template.addRange(range, standardHtmlEvent);
             }
           }
+
+          //TODO: Refactor the following chunk of statement resolver
+          List<Statement> statements =
+              _resolveDartStatementsAt(valueOffset, value, eventType);
+          for (Statement statement in statements) {
+            _recordAstNodeResolvedRanges(statement);
```
MichaelRFairhurst: Doesn't this already happen in `_resolveDartStatementsAt`?

analyzer_plugin/test/resolver_test.dart
```diff
@@ -822,6 +822,96 @@ class GenericComponent<T extends String> {
...
+''');
+    String code = r"""
+<div (click)='handleClick($event); unknownFunction()'></div>
+""";
+    _addHtmlSource(code);
+    _resolveSingleTemplate(dartSource);
+    assertErrorInCodeAtPosition(
+        StaticTypeWarningCode.UNDEFINED_METHOD, code, "unknownFunction");
+  }
+
```
MichaelRFairhurst: Missing test cases:

```
(click)="if(x) { }"
(click)="return x"
(click)="x;}"
```

The first two are statements which the parser will return but don't belong in angular templates. The second one will probably have the same effect that `[x]="x;}"` had before your latest fixes, where the parser returns without consuming all of the input tokens.

MichaelRFairhurst:
Wow, big harsh red x! First time reviewing, excuse me!

Looks good Max!! I'm super glad to have you taking the helm on this stuff. Very detail oriented stuff, no?

I hope the remaining items on here aren't too much extra work, though I do have a feeling some of them might get a bit hairy, if so we can merge anything that doesn't cause false errors, and worry about that extra bit of completeness later. I think the only thing in here we really _shouldn't_ merge is the double resolveAst if I followed the code right, and removing code that's no longer used if it no longer is. The rest is just for the sake of completeness!

analyzer_plugin/lib/src/resolver.dart
mk13: _resolveDartStatementsAt goes through each statement and calls _resolveDartAstNode, but doesn't call _recordAstNodeResolvedRanges.

analyzer_plugin/lib/src/resolver.dart
mk13: Removing them shortly.

analyzer_plugin/lib/src/resolver.dart
mk13: It should no longer be in the file and shouldn't used anywhere. For all intents and purposes, _recordAstNodeResolvedRanges now completely replaces _recordExpressionResolvedRanges.

analyzer_plugin/lib/src/resolver.dart
MichaelRFairhurst: Ah! Ok great. No problem here, my bad

analyzer_plugin/test/resolver_test.dart
mk13: They'll be added shortly.

analyzer_plugin/lib/src/resolver.dart
MichaelRFairhurst: Yes. Once again, my bad here. You did remove it, I didn't notice the part where you had.

mk13:
No problem - thanks for the overview. It's a pretty big change and I'd rather play this on the safe side. 

I'll finalize the additional requested test cases, and we should definitely talk about this tomorrow to go through all potential side-effects (ex: false errors).

MichaelRFairhurst:
Approved it, ideally we can still put in & handle those extra test cases but as-is it causes no false errors and catches more than the code used to support.

Thanks Max! Sorry again for the fail first time review....

Issue #62, opened 2016-11-17
Error for ngFor not in a template


MichaelRFairhurst:
If I write

```<div ngFor="let x of xs">...</div>```

I need to see an error reminding me to use `*ngFor`.

Can also occur with

```<div ngFor let-x [ngForOf]="xs">...</div>```

in which case the proper solution is to put `<template ngFor let-x [ngForOf]="xs"><div>...</div></template>` however this is fringe.

These could be detected separately. In the former case, the ngFor directive will not match the element, where in the former it will. Reporting an error for unknown properties could solve this.

In the latter case, the ngFor directive is matched but its not in a template element.

If caught separately, we could report errors.

But its probably fine just to check node.attributes for "ngFor", and if `!node.isOrHasTemplateAttribute` then report a single error which suggests both solutions in the error message.

MichaelRFairhurst:
Merged #113

PR #113, opened 2016-12-01
Check that ngIf and ngFor are used in a template refs #62

MichaelRFairhurst:


analyzer_plugin/lib/tasks.dart
```diff
@@ -239,6 +239,14 @@ class AngularWarningCode extends ErrorCode {
               'or class binding');
 
   /**
+   * An error code indicating that ngIf or ngFor were used without a template
+   */
+  static const AngularWarningCode STRUCTURAL_DIRECTIVES_REQUIRE_TEMPLATE =
+      const AngularWarningCode('STRUCTURAL_DIRECTIVES_REQUIRE_TEMPLATE',
+          'Structural directive {0} requires a template. Did you mean ' +
+              '*{0}="..." or template="{0} ..." or <template {0} ...>?');
```
scheglov: FYI, [AngularWarningCode] constructor has the third parameter `correction`.
But you're free to use message if you think this will provide better UX.

MichaelRFairhurst: Interesting...what exactly does it do? Tried using it but it looks like it just disappeared. Does it hook into the suggestion API or something?

scheglov: Well, `correction` is sent to the client of Analysis Server.
But it seems that IDEA does not show it :-(
So, I think you don't want to use it, at least not now ;-)

PR #118, opened 2016-12-02
Refactor resolve attribute value

mk13:
Refactored _resolveAttributeValues by splitting use-cases for input, output, and two-way bound attribute values. 

Things to note: 'eventType' parameter in _resolveDartExpressionAt(..) function (and subsequent functions) is removed; 'eventType' only has a value during output-bound attribute values - which is handled completely now by _resolveDartStatementAt(...).


analyzer_plugin/lib/src/resolver.dart
```diff
@@ -865,6 +761,145 @@ class TemplateResolver {
   }
 
   /**
+   * Resolve output-bound values of [attributes] as statements.
+   */
+  void _resolveOutputBoundAttributeValues(int valueOffset, String value,
```
MichaelRFairhurst: `value` and `valueOffset` are just `attribute.value` and `attribute.valueOffset`, probably can leave them out the function signature

analyzer_plugin/lib/src/resolver.dart
```diff
@@ -712,109 +712,37 @@ class TemplateResolver {
...
       String value = attribute.value;
-      dart.DartType eventType = null;
+
       // already handled
       if (attribute.name == 'template' || attribute.name.startsWith('*')) {
         continue;
       }
+
       // bound
       if (attribute.bound != null) {
```
MichaelRFairhurst: We could probably change this to

```
if (attribute.bound == AttributeBoundType.output) {
  ...
} else if (attribute.bound != null) {
  ...
} else {
}
```

and drop the 'continue's.

Small simple idea

analyzer_plugin/lib/src/resolver.dart
mk13: Changing it shortly.

PR #131, opened 2016-12-06
Statement requires exp stmt

mk13:
Moved 'ExpressionStatement' check to _resolveDartExpressionStatementsAt(..) to occur before '_resolveAstNode(..)'; this is because passing a non-ExpressionStatement into '_resolveAstNode(..)' is causing errors deep within the sdk stack. This prevents the issue from arising.


analyzer_plugin/test/resolver_test.dart
```diff
@@ -912,21 +913,80 @@ class TestPanel {
     errorListener.assertNoErrors();
   }
 
-  void test_statement_eventBinding_return_statement() {
+  void test_statement_eventBinding_return_statement_without_semicolon() {
     _addDartSource(r'''
-import 'dart:html';
+import 'dart:html'
```
MichaelRFairhurst: is this removed semicolon intentional?

analyzer_plugin/test/resolver_test.dart
```diff
@@ -912,21 +913,80 @@ class TestPanel {
...
-    errorListener.assertNoErrors();
+    assertErrorInCodeAtPosition(
+        AngularWarningCode.OUTPUT_STATEMENT_REQUIRES_EXPRESSION_STATEMENT,
+        code,
+        "return 5");
+  }
+
+  void test_statement_eventBinding_return_statement_with_semicolon() {
+    _addDartSource(r'''
+import 'dart:html'
```
MichaelRFairhurst: also missing here

analyzer_plugin/test/resolver_test.dart
```diff
@@ -966,6 +1026,138 @@ class TestPanel {
...
+''');
+    String code = r"""
+<div (click)='handleClick;'></div>
+    """;
+    _addHtmlSource(code);
+    _resolveSingleTemplate(dartSource);
+    errorListener.assertNoErrors();
+  }
+
+  void test_statement_eventBinding_single_variable2() {
```
MichaelRFairhurst: Definitely nice to err on the side of overprepared, but I don't think we need two tests for a single variable, one should probably do

analyzer_plugin/test/resolver_test.dart
```diff
@@ -966,6 +1026,138 @@ class TestPanel {
...
+import 'dart:html';
+@Component(selector: 'test-panel')
+@View(templateUrl: 'test_panel.html')
+class TestPanel {
+  void handleClick(MouseEvent e) {
+  }
+}
+''');
+    String code = r"""
+<div (click)='handleClick($event)};'></div>
```
MichaelRFairhurst: This isn't the right test case.

The grammar for an expression statement is: an expression followed by a semicolon. `handleClick($event)` is an expression, so testing that it catches the unexpected '}' is really just testing the parser, not our own code.

On the other hand, if there were a semicolon, ie, `handleClick($event);}`' then the statements parse step will probably see the `}` simply as a signal that its done, and ignore from there, just like the bug we had recently with trailing expressions.

analyzer_plugin/lib/tasks.dart
```diff
@@ -255,6 +259,15 @@ class AngularWarningCode extends ErrorCode {
           'No directives matching this element are exported by the name {0}');
 
   /**
+   * An error code indicating that an output-bound statement
+   * must be an [ExpressionStatement].
+   */
+  static const AngularWarningCode
+      OUTPUT_STATEMENT_REQUIRES_EXPRESSION_STATEMENT = const AngularWarningCode(
+          'OUTPUT_STATEMENT_REQUIRES_EXPRESSION_STATEMENT',
+          'An output-bound statement must be an ExpressionStatement');
```
MichaelRFairhurst: I've got an idea on how this warning message could be clearer.

My first thought is that people probably won't know what 'expression statement' means, so it might be better to say "return statement is not allowed", but I'm not sure the best way to get the name of a node.

Then I realized that it might be simpler to just say "syntax error: unexpected token IF" and then regardless of the type of statement people should realize the type they tried was not allowed. It also should feel familiar since that's how most parsers work.

Either way, you can embed parameters in these error messages by using `{0}`, `{1}`, etc, and then passing the data in when you report the error ie `.onError(new AnalysisError(src, offs, len, code, [item0, item1]))`

analyzer_plugin/lib/src/resolver.dart
```diff
@@ -649,10 +649,38 @@ class TemplateResolver {
...
+          while (token.type == TokenType.CLOSE_CURLY_BRACKET) {
+            token = token.next;
+          }
+          int length = token.offset - startCloseBracket;
+          errorListener.onError(new AnalysisError(
+              templateSource,
+              startCloseBracket,
+              length,
+              ParserErrorCode.UNEXPECTED_TOKEN,
+              ["}" * length]));
```
MichaelRFairhurst: Technicality here: "}}" isn't a token, only "}" is. You can probably just report "}" and skip over to the next non "}" and resume parsing

analyzer_plugin/lib/src/resolver.dart
```diff
@@ -649,10 +649,38 @@ class TemplateResolver {
...
+          continue;
+        }
+        //Nothing should trigger here, but just in case to prevent infinite loop
+        token = token.next;
```
MichaelRFairhurst: good catchall

analyzer_plugin/lib/src/resolver.dart
```diff
@@ -1050,26 +1078,59 @@ class TemplateResolver {
...
+        _resolveDartAstNode(statement, eventType);
       }
     }
     return statements;
   }
 
   /**
+   * Get helpful description based on statement type to report in
+   * OUTPUT_STATEMENT_REQUIRES_EXPRESSION_STATEMENT
+   */
```
MichaelRFairhurst: What about `new AnalysisError(templateSource, offset, length, PARSE_ERROR, [stmt.token.type])`? Does that look clean?

analyzer_plugin/test/resolver_test.dart
```diff
@@ -966,6 +1026,119 @@ class TestPanel {
...
+''');
+    String code = r"""
+<div (click)='handleClick($event);}}}}'></div>
+    """;
+    _addHtmlSource(code);
+    _resolveSingleTemplate(dartSource);
+    assertErrorInCodeAtPosition(ParserErrorCode.UNEXPECTED_TOKEN, code, '}}}}');
+  }
+
+  void test_statement_eventBinding_unexpected_closing_brackets_at_start() {
```
MichaelRFairhurst: Might be nice to add a test that code after "}" is still typechecked, ie, check that "}1.length" reports an unknown getter, or some kind of test like that

analyzer_plugin/lib/src/resolver.dart
mk13: Sort of. That was the second solution - just to state "Syntax error: unexpected <insert statement type>"

analyzer_plugin/lib/src/resolver.dart
mk13: Sure, I will remove the "* length" so it will just say "} is an unexpected token"

analyzer_plugin/lib/src/resolver.dart
```diff
@@ -649,10 +649,38 @@ class TemplateResolver {
...
-    return _parseDartStatementsAtToken(token);
+
+    while (token.type != TokenType.EOF) {
+      List<Statement> currentStatements = _parseDartStatementsAtToken(token);
+
+      if (currentStatements.isNotEmpty) {
+        allStatements.addAll(currentStatements);
+        token = currentStatements.last.endToken.next;
+      } else {
+        if (token.type == TokenType.CLOSE_CURLY_BRACKET) {
```
MichaelRFairhurst: Hmm, rereading this code I'm not sure 'else' is best here. The contents of the else block finds, reports and skips '}'. That doesn't have anything to do with whether the current statements are empty, does it?

does
```
if (currentStatements.isNotEmpty) {
  // move `token`
  ...
}

if (token.type == TokenType.CLOSE_CURLY_BRACKET) {
  ...
```

still pass the test cases?

MichaelRFairhurst:
I say merge this when you're ready even though the tests are failing. I will try to fix it soon now that I've got autocomplete in a reviewable stage by the masters of dart analysis things.

PR #143, opened 2016-12-07
Preliminary autocompletion work for review

MichaelRFairhurst:
@scheglov if you can review this before I go too far, that would be great!

Copy some test code from analysis server to be able to write
autocompletion tests. Basic plugin to autocomplete and dart
autocomplete (which are separate pluggable things).

Store expressions in templates with their offsets so that we can get the
ASTs after the fact for autocomplete.

So far just returns a hardcoded result 'text' for anything that looks
like its completing an angular expression in dart code.

### DIFF
--- a/analyzer_plugin/lib/src/model.dart
+++ b/analyzer_plugin/lib/src/model.dart
@@ -2,6 +2,7 @@ library angular2.src.analysis.analyzer_plugin.src.model;
 
 import 'package:analyzer/dart/element/element.dart' as dart;
 import 'package:analyzer/dart/element/type.dart' as dart;
+import 'package:analyzer/dart/ast/ast.dart' as dart;
 import 'package:analyzer/src/generated/source.dart' show Source, SourceRange;
 import 'package:analyzer/src/generated/utilities_general.dart';
 import 'package:analyzer/task/model.dart' show AnalysisTarget;
@@ -243,6 +244,25 @@ class ResolvedRange {
   }
 }
 
+class EmbeddedExpression {
+  /**
+   * The [SourceRange] where [element] is referenced.
+   */
+  final SourceRange range;
+
+  /**
+   * The [Expression] referenced at [range].
+   */
+  final dart.Expression expression;
+
+  EmbeddedExpression(this.range, this.expression);
+
+  @override
+  String toString() {
+    return '$range=[$expression]';
+  }
+}
+
 /**
  * An Angular template.
  * Templates can be embedded into Dart.
@@ -265,6 +285,11 @@ class Template {
    */
   final List<ResolvedRange> ranges = <ResolvedRange>[];
 
+  /**
+   * The [EmbeddedExpression]s of the template.
+   */
+  final List<EmbeddedExpression> embeddedExpressions = <EmbeddedExpression>[];
+
   Template(this.view, this.element);
 
   /**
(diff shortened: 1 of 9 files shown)


analyzer_plugin/lib/src/model.dart
```diff
@@ -243,6 +244,25 @@ class ResolvedRange {
   }
 }
 
+class EmbeddedExpression {
+  /**
+   * The [SourceRange] where [element] is referenced.
+   */
+  final SourceRange range;
```
scheglov: Element, what element?
In theory Expression already knows its source range: [`expression.offset`, `expression.end`).

scheglov:
@bwilkerson

server_plugin/lib/src/analysis.dart
```diff
@@ -236,3 +252,55 @@ class AngularOccurrencesContributor implements OccurrencesContributor {
...
+    return suggestions;
+  }
+}
+
+/**
+ * An object used to produce completions at a specific location within a file.
+ *
+ * Clients may implement this class when implementing plugins.
+ */
+class AngularTemplateCompletionContributor extends CompletionContributor {
```
scheglov: The comment is not relevant here.

server_plugin/lib/src/analysis.dart
```diff
@@ -236,3 +252,55 @@ class AngularOccurrencesContributor implements OccurrencesContributor {
...
+  /**
+   * Return a [Future] that completes with a list of suggestions
+   * for the given completion [request]. This will
+   * throw [AbortCompletion] if the completion request has been aborted.
+   */
+  Future<List<CompletionSuggestion>> computeSuggestions(
+      CompletionRequest request) {
+    return new Future.value([]);
```
scheglov: Make it `async`.
Return `[]`.

server_plugin/test/completion_contributor_test.dart
```diff
@@ -0,0 +1,36 @@
+library angular2.src.analysis.server_plugin.analysis_test;
```
scheglov: Remove `library` directive everywhere.

server_plugin/lib/src/analysis.dart
MichaelRFairhurst: Good catch

MichaelRFairhurst:
Hmm, wouldn't this be more or less equally applicable to autocompleting html files?

Or do you mean, like, autocompleting html in html files...because, yes, that information is somewhat lost. I can still grab the `Document` that's parsed to do regular AST traversals for all that normal stuff.

But I do have an issue of, for instance, autocompleting ng-for items since they are scoped in that tree in special rules...

So instead then I should probably create my own AST (or more like just expand NodeInfo and AttributeInfo) to detail things like variables in scope for each expression and what directives are on the element etc. Does that sound like a better approach?

scheglov:
Yes, that sounds reasonable.

To complete top-level things in expressions (i.e. not after a `.`) you basically need the same scopes that you would use to resolve these things.

MichaelRFairhurst:
New changes that could use more review from @scheglov @bwilkerson:

PR for enabling work in sdk https://codereview.chromium.org/2565633002

Instantiate and invoke the some dart completers directly: inheritance
contributor which gathers members based on the enclosing class of the
template, and typed member contributor.

Using these contributors requires a DartCompletionRequest, and the impl
does a lot of things with CompilationUnits which we don't have. So we
implement the class ourselves and shut out the things we don't care
about. Some code duplication here unfortunately.

PR #144, opened 2016-12-07
Change ErrorVerifier API to use latest

MichaelRFairhurst:


MichaelRFairhurst:
maybe @keertip can help answer a question I have here...

If I merge this, will this delay release to google3 users until we say, release 1.21.dev-25 or something like that?

Maybe we shouldn't be building against the SDK at HEAD?

keertip:
If you want the angular analyzer plugin released along with the server in the internal SDK, then you should be building at HEAD. That is what will get pulled in, though only the dev builds.

This also means that the angular plugin release cycle will get tied in with the Dart SDK, and  iterating quickly may not be possible.

PR #153, opened 2016-12-10
Use sdk packages file

MichaelRFairhurst:
Scripted way of downloading deps, and rolling them up into .packages.

Why am I doing this? This feels absolutely horrible and I wish I could
just use pub.


MichaelRFairhurst:
YES fixed travis. And I think this is a much less fragile solution...though I still wish we were just using pub in a reliable way.

mk13:
Looks great - working well :)

PR #157, opened 2016-12-13
'hash' vs 'let' usage in templates #refs 53

mk13:
Properly flag improper use of '#' while still treating the variable prefixed with '#' as a valid variable to avoid unnecessary flags being raised.

For example:
```
<div *ngFor="#item of items">
    {{ item.name }}
</div>
```
'item' token will now be known by the parser and won't flag it as an unknown variable within the structural directive. The only error will be on the '#' token.


mk13:
Let me know if you want to hold off on this merge until the refactoring of resolveTemplateAttribute is complete.

PR #158, opened 2016-12-14
Check expressions in unmatched mustaches #refs 146

mk13:
Expressions on unclosed moustaches will now be parsed like a standard expression until end of text or next error-unclosed moustache is found.


MichaelRFairhurst:
What are your thoughts about when people put these mustaches in the middle of prior text?

ie existing file shows something to the effect of
```
  <div>Welcome from all of us at mysite.com! Keep coming back for rewards!</div>
```

and then they go to change mysite.com into a bound expression:

```
  <div>Welcome from all of us at {{site_url! Keep coming back for rewards!</div>
                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

site_url autocompletes for them, and now the sentence afterwards is highlighted as a trailing error expression.

We could suppress that warning, but on the other hand, the fix is pretty easy (just type in `}}`). Also not sure how common that workflow would be.

What are your thoughts there?

mk13:
If the user types in "{{" but never closes it - do we have to suppress errors following it? Technically the user made a mistake by starting a bound expression but never closing it. If I'm not mistaken, the "{{" should also raise a flag giving the user a easy hint as to why the following sentence is being also flagged as well.

MichaelRFairhurst:
I guess my point is if they have typed in `{{valid_expr not valid expr here` should we be flagging 'not valid expr here` or giving them the benefit of the doubt that its just text after the expression?

I think in general the rule of not showing fallout errors applies. As in, if we know the `{{` is unclosed, we shouldn't report trailing expression because they could be plain text, and they already have the unclosed error. If they then close it around the trailing text, then we can report the trailing text.

But I'm wondering what the likelihood really is

mk13:
Hmmm sure. That sounds like a reasonable assumption that if the expression ends within the string of text, the user meant the beyond-eof text to be plain text rather than an expression. 

It should be a simple fix to only parse as much expression as it sees following {{ and ignore the rest following the eof. Once the }} is found, it can then complain about the non-valid trailing expression portion.

PR #159, opened 2016-12-15
Create an AST for angular templates and resolve with visitors

MichaelRFairhurst:
The AST is mostly just the ElementInfo and AttributeInfo classes from
before, but with more subclassing involved now (template, text,
expression-bound, and statement-bound attributes split up). Also a node
for mustaches, and the local scope stored on the tree as well for
autocomplete support.

Move the html-ast to angular-ast conversion into converter.dart, which
parser the dart up front, as well as template attributes up front, just
doesn't resolve the dart expressions/statements it builds yet.

Once the conversion is done, break the resolution up into three
visitors:
1. straight shot directive resolver. Since template attributes are
broken down already in the AST, we can resolve directives on the
elemenst and the sugared templates in one straight go.
2. Scope preparer. The parent scope has to be resolved before the next
one can be resolved, so it prepares just one scope. Loads all #vars (as
well as ngFor vars from the template below), and store the corresponding
scopes in the AST so they're ready to be resolved. Resolve outputs so
that $event vars can be loaded in as well.
3. Scope resolver. Once the ASTs are prepared with scopes, resolve them
all. Pretty simple.

Jumping the gun to changing the resolution to using the visitors will
help us transition to the other angular parser which we will one day use
which also relies on an AST and visitors.

Could still use some work, specifically in the complexity of how scopes
are chained. The scope preparer and scope resolver kind of share the
same work there, although it's not immediately obvious how to best
abstract that.

Also made the scope preparer load event variables even though its a very
isolated second pass which sholudn't really be done in the same visitor,
but the code to create variables (which $event vars are) is nontrivial,
and this way it gets to utilize the scoping logic of the scope preparer
which is also, as mentioned before, nontrivial and needing an
abstraction.

That said, I think its a nice step for the code in general and will
serve autocompletion well...or at least, its close to a nice change in
the code. Up to the reviewers to decide that I guess.

### DIFF
--- a/analyzer_plugin/lib/ast.dart
+++ b/analyzer_plugin/lib/ast.dart
@@ -20,8 +20,7 @@ abstract class AngularAstVisitor {
 
   void visitTextAttr(TextAttribute textAttr) => _visitAllChildren(textAttr);
 
-  void visitTemplateAttr(TemplateAttribute attr) =>
-      _visitAllChildren(attr);
+  void visitTemplateAttr(TemplateAttribute attr) => _visitAllChildren(attr);
 
   void visitExpressionBoundAttr(ExpressionBoundAttribute attr) =>
       _visitAllChildren(attr);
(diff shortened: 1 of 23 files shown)


analyzer_plugin/lib/ast.dart
```diff
@@ -0,0 +1,251 @@
...
+class TextInfo extends NodeInfo {
+  final List<Mustache> mustaches;
+  List<AngularAstNode> get children => new List<AngularAstNode>.from(mustaches);
+
+  final String text;
+  final int offset;
+
+  TextInfo(this.offset, this.text, this.mustaches);
+
+  int get length => offset + text.length;
```
mk13: I'm not sure here: why is the length equivalent to the end offset of the string text?

MichaelRFairhurst: Ah, yes, good catch! This should just be text.length

MichaelRFairhurst: In general some of the offset stuff is off. At the moment, it only matters for the autocomplete branch, so I've been fixing it in there. Looks like this was fixed in that branch

MichaelRFairhurst:
Awesome! Was going to merge, but I'm seeing some issues with half-complete html crashing the analyzer. Pushed the fix. Going to manually test this before merging, we only *need* to merge it to launch autocomplete

MichaelRFairhurst:
Looks like this branch should be merged sooner rather than later for two issues it solves:

* `[input]` and `[input]=""` and `(output)` (though not `(output)=""`, still feels valid to me...) errors. 
* `attr=""` crashes pre-this branch (hey a bug that the refactor solved unknowingly, instead of causing unknowingly! That's always a good sign)

Might call for another dogfood release though, I'm a little scared of how much has changed and what new bugs it may have caused.

MichaelRFairhurst:
Closing this, not merging. #179 includes these changes, but since I did a squash merge, it technically doesn't include these commits so github still thinks this needs to be opened.

PR #160, opened 2016-12-16
Remove false errors on pipes #refs 155

mk13:
Treat expressions with '|' as an 'asExpression' instead of as a 'bitwiseOrExpression'. Extend existing Dart parser to override the 'parseBitwiseOrExpression' to properly resolve piped expressions. 

Left side of pipe should be properly checked by the resolver while anything to the right of the pipe are ignored and will not raise any errors.


analyzer_plugin/lib/src/ng_expr_parser.dart
```diff
@@ -0,0 +1,71 @@
+library angular2.src.analysis.analyzer_plugin.src.ng_expr_parser.dart;
```
MichaelRFairhurst: You can actually remove these, they don't do anything....

analyzer_plugin/lib/src/ng_expr_parser.dart
```diff
@@ -0,0 +1,71 @@
...
+import 'package:analyzer/src/generated/resolver.dart';
+import 'package:analyzer/dart/ast/standard_ast_factory.dart';
+import 'package:front_end/src/scanner/token.dart';
+
+class NgExprParser extends Parser {
+  NgExprParser(
+      Source source, AnalysisErrorListener errorListener, this.typeProvider)
+      : super(source, errorListener);
+
+  Token get _currentToken => super.currentToken;
```
MichaelRFairhurst: Hmm, is this necessary?

analyzer_plugin/lib/src/ng_expr_parser.dart
```diff
@@ -0,0 +1,71 @@
...
+
+  final TypeProvider typeProvider;
+
+  @override
```
MichaelRFairhurst: dangling annotation

analyzer_plugin/lib/src/ng_expr_parser.dart
```diff
@@ -0,0 +1,71 @@
...
+   * Parse a bitwise or expression to be treated as a pipe.
+   * Return the resolved left-hand expression as a dynamic type.
+   *
+   *     bitwiseOrExpression ::=
+   *         bitwiseXorExpression ('|' pipeExpression)*
+   */
+  @override
+  Expression parseBitwiseOrExpression() {
+    Expression expression;
+    Token tempEndToken = null;
```
MichaelRFairhurst: maybe could use a name like 'beforePipeToken`

analyzer_plugin/lib/src/ng_expr_parser.dart
```diff
@@ -0,0 +1,71 @@
...
+    return expression;
+  }
+
+  /**
+   * Parse a bitwise or expression to be treated as a pipe.
+   * Return the resolved left-hand expression as a dynamic type.
+   *
+   *     pipeExpression ::= identifier[':' expression]*
+   */
+  void parsePipeExpression() {
```
MichaelRFairhurst: Very clean, very nice

analyzer_plugin/test/resolver_test.dart
```diff
@@ -877,6 +877,75 @@ class GenericComponent<T extends String> {
...
+''');
+    _addHtmlSource(r"""
+<li *ngFor='let operator of (operators | pipe1)'>
+  {{operator.length}}
+</li>
+""");
+    _resolveSingleTemplate(dartSource);
+    errorListener.assertNoErrors();
+  }
+
```
MichaelRFairhurst: Some of the edge cases you found would be good to have as test cases, such as (a | number) + (b | number)

analyzer_plugin/lib/src/ng_expr_parser.dart
mk13: It was flagging _currentToken as not found; it would probably work without it though. I can remove it though.

analyzer_plugin/test/resolver_test.dart
mk13: Used my long gnarly parathensized expression chain instead.

mk13:
Cool, changes have been made. And yea, I think it would be ready for next sdk release.

analyzer_plugin/lib/src/ng_expr_parser.dart
MichaelRFairhurst: hmm, but shouldn't you be able to just do 'currentToken'?

PR #164, opened 2016-12-20
Handle sanest and some of the crazier constant template expressions

MichaelRFairhurst:
PR will fail to build until https://codereview.chromium.org/2586363004/ is merged.

Handle template: 'a' + 'b', etc.

Use ConstantEvaluator for most of the heavy lifting, but for things like
templates where we need to track offsets, use a special new visitor
(with tests) to pad in appropriately to match the constant expression
around operators, parens etc.

recognize when adding spaces can't fix stuff, and bail out.

### DIFF
--- a/analyzer_plugin/lib/src/tasks.dart
+++ b/analyzer_plugin/lib/src/tasks.dart
@@ -2,6 +2,7 @@ library angular2.src.analysis.analyzer_plugin.src.tasks;
 
 import 'package:analyzer/src/context/cache.dart';
 import 'package:analyzer/dart/ast/ast.dart' as ast;
+import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:analyzer/src/dart/ast/utilities.dart' as utils;
 import 'package:analyzer/src/generated/constant.dart';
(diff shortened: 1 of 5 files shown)


MichaelRFairhurst:
Build should be fixed now

analyzer_plugin/lib/src/tasks.dart
```diff
@@ -658,17 +661,16 @@ class BuildUnitDirectivesTask extends SourceBasedAnalysisTask
...
-    String selectorStr;
-    int selectorOffset;
-    if (expression is ast.SimpleStringLiteral) {
-      selectorStr = expression.value;
-      selectorOffset = expression.contentsOffset;
-    } else {
-      errorReporter.reportErrorForNode(
-          AngularWarningCode.STRING_VALUE_EXPECTED, expression);
+    // Compute the selector text. Careful! Offsets may not be valid after this,
+    // however, at the moment we don't use them anyway.
```
mk13: In future cases where strict offsets are needed, would it be too much of a runtime penalty to store the expression within the selectors? It would make offset calculation easier after this step finishes.

analyzer_plugin/lib/src/tasks.dart
```diff
@@ -1306,6 +1311,113 @@ class ResolveHtmlTemplateTask extends AnalysisTask {
   }
 }
 
+class OffsettingConstantEvaluator extends utils.ConstantEvaluator {
```
mk13: Interesting solution to preserve offsets!

PR #179, opened 2016-12-21
Autocomplete ast

MichaelRFairhurst:
This will be moreso the master branch for autocompletion I think.

### DIFF
--- a/analyzer_plugin/lib/src/converter.dart
+++ b/analyzer_plugin/lib/src/converter.dart
@@ -168,7 +168,7 @@ class HtmlTreeConverter {
         valueOffset,
         origName,
         origNameOffset,
-        dartParser.parseDartExpression(valueOffset, value),
+        dartParser.parseDartExpression(valueOffset, value, true),
         bound);
   }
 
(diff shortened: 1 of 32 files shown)


analyzer_plugin/lib/src/converter.dart
```diff
@@ -0,0 +1,558 @@
...
+
+      if (end == -1) {
+        errorListener.onError(new AnalysisError(templateSource,
+            fileOffset + begin, 2, AngularWarningCode.UNTERMINATED_MUSTACHE));
+        // Move the cursor ahead and keep looking for more unmatched mustaches.
+        textOffset = begin + 2;
+        exprBegin = textOffset;
+        exprEnd = _startsWithWhitespace(text.substring(exprBegin))
+            ? exprBegin
+            : text.length;
```
mk13: This would make 'code' into an empty string and expression to hold an empty expression. Wouldn't this break error detection in unterminated mustaches?

mk13:
Interesting implementation of the autocomplete visitor. It looks good overall and eager to get all the visitor PRs implemented.

PR #181, opened 2016-12-22
Method to suppress analysis warnings

MichaelRFairhurst:


analyzer_plugin/test/tasks_test.dart
```diff
@@ -1636,6 +1636,77 @@ class ComponentA {
         AngularWarningCode.UNRESOLVED_TAG, code, 'unresolved-tag');
   }
 
+  void test_suppressError_UnresolvedTag() {
+    String code = r'''
+import '/angular2/angular2.dart';
+
+@Component(selector: 'my-aaa',
+    template: """
+<!-- @ngIgnoreErrors: UNRESOLVED_TAG -->
```
zoechi: Why not stick with the way the analyzer/linter suppress errors/warnings?
```dart
// ignore: overridden_fields, other_warning
```
```html
<!-- ignore: unresolved_tag, other_warning -->
```
or
```html
<!-- ignore: ng_unresolved_tag, ng_other_warning -->
```

They recently added support for suppressing by adding comments at the end of the line or just before the line with the error/warning.

MichaelRFairhurst: I definitely wanted something more specific than just "ignore," though this could be a good case for lowercase (or case insensitive) error code names.

Its conceivable that someone could use multiple html extensions, perhaps for instance as a pub transformer, and that the extended html could come with its own analyzer plugin with its own "unresolved_tag" error etc.  Or copying markup around where the comment shouldn't have an effect, and should be clearly an angular feature.

Pretty stretch concern, for sure. But just in general I lean more towards the java `@SuppressWarnings(...)` clarity over the dart `ignore:` simplicity.

Based on that reasoning, I do like the NG prefix on the error codes. One thought off of that, perhaps positive, perhaps negative, is we would have the option to differentiate clearly between dart errors within ng templates and ng errors. For instance, `UNKNOWN_IDENTIFIER` is raised by only indirectly by our plugin, and is a standard dart error, as opposed to `UNRECOGNIZED_TAG` which is our error code raised by us. I'm trying to decide whether mixing ng-prefixed things like `NG_INPUT_NOT_RECOGNIZED` with nonprefixed things like `UNKNOWN_GETTER` is a hair clearer or a hair foggier.

If error codes were prefixed, prefixing the 'ignore' declaration at the top might not be necessary. This could be a cool thing for mixing different types of html validation, and sharing an ignore tag at the top. And if they both raise dart errors, leaving those in unprefixed could mean those get shared too.

Overall, this is just a stopgap fix. I'll add case insensitivity but leave the prefixes as they are now.

Issue #183, opened 2016-12-22
Analysis crash putting @Input on a final var


MichaelRFairhurst:
Turns out not all `FieldDeclaration`s have setters, which the code assumes. Need to raise an error when that happens.

```
   @Input()
   final bool isEnabled = true;
```

This is also clearly not an intended state of the code, as you can't use this input since `isEnabled` is immutable.

MichaelRFairhurst:
#186 merged

PR #186, opened 2016-12-23
When a declared (in|out)put is malformed, skip it to prevent crashes …

MichaelRFairhurst:
…refs #183

Used to create a malformed InputElement (or OutputElement) that was
added to the directive and crashed the server when used since it was in
a half complete state.

We could set it to something dynamic, but references won't be traceable
and it should be an error to use an erroneous input since that
erroneous input may have been written by someone else without analysis
and you would have a bad day if you were allowed to use such an input,
discussion in #183 on github.


mk13:
> We could set it to something dynamic, but references won't be traceable
and it should be an error to use an erroneous input since that
erroneous input may have been written by someone else without analysis
and you would have a bad day if you were allowed to use such an input,
discussion in #183 on github.

I think it's fine to state the immutable input field as "not found" - which this and the other @Input fix would continue to do. In future fixes, we can potentially provide better warnings.

Issue #120, opened 2016-12-02
Exception from navigation contributor: AngularNavigationContributor. The getter 'offset' was called on null.


alexander-doroshko:
## Analyzer Feedback from IntelliJ
IDEA: 171.SNAPSHOT, Dart: 1.21.0-dev.11.1, Windows 10
## Stack

```dart
Dart analysis server, SDK version 1.21.0-dev.11.1, server version 1.17.0, error: Exception from navigation contributor: AngularNavigationContributor
NoSuchMethodError: The getter 'offset' was called on null.
Receiver: null
Tried calling: offset
#0      Object._noSuchMethod (dart:core-patch/object_patch.dart:44)
#1      Object.noSuchMethod (dart:core-patch/object_patch.dart:47)
#2      AngularNavigationContributor._addDirectiveRegions (package:angular_analyzer_server_plugin/src/analysis.dart:94)
#3      AngularNavigationContributor.computeNavigation (package:angular_analyzer_server_plugin/src/analysis.dart:54)
#4      computeNavigation (package:analysis_server/src/domains/analysis/navigation.dart:28)
#5      sendAnalysisNotificationNavigation.<anonymous closure> (package:analysis_server/src/operation/operation_analysis.dart:203)
#6      _sendNotification.<anonymous closure> (package:analysis_server/src/operation/operation_analysis.dart:274)
#7      _PerformanceTagImpl.makeCurrentWhile (package:analyzer/src/generated/utilities_general.dart:189)
#8      _sendNotification (package:analysis_server/src/operation/operation_analysis.dart:272)
#9      sendAnalysisNotificationNavigation (package:analysis_server/src/operation/operation_analysis.dart:201)
#10     AnalysisServer.setAnalysisSubscriptions.<anonymous closure> (package:analysis_server/src/analysis_server.dart:1242)
#11     _HashMap.forEach (dart:collection-patch/collection_patch.dart:151)
#12     AnalysisServer.setAnalysisSubscriptions (package:analysis_server/src/analysis_server.dart:1219)
#13     AnalysisDomainHandler.setSubscriptions (package:analysis_server/src/domain_analysis.dart:307)
#14     AnalysisDomainHandler.handleRequest (package:analysis_server/src/domain_analysis.dart:220)
#15     AnalysisServer.handleRequest.<anonymous closure>.<anonymous closure> (package:analysis_server/src/analysis_server.dart:864)
[...]

PR #187, opened 2016-12-23
Fix null pointer in navigation contributor refs #120

MichaelRFairhurst:
problem was null 'setterRange' and 'getterRange' when using `@Input` and `@Output`


MichaelRFairhurst:
I think the analysis failure in this branch is solved in a different one. Not sure which one though, its caused by an sdk change unrelated to anything we've done recently, I think I solved it in a random branch....

MichaelRFairhurst:
Ping @mk13, I finally fixed the build here

PR #188, opened 2016-12-28
Autocomplete outputs within other outputs, and standard events.

MichaelRFairhurst:
Suggest outputs in the middle of writing them.

Suggest events like click, mouseover, etc, all at lower priority than
other directive outputs, though ideally they would be at the same
priority on standard dom. Could adjust that later.

Tests that they are suggested at the right times at the right priority.


MichaelRFairhurst:
More goodies thrown in:

    Don't autocomplete the same input or output twice, with tests.
    
    Use hashsets for better performance.
    
    Also make sure inputs and outputs still suggest themselves (for
    instance, you may have both (click) and (clickTwice) events, and want
    to start completing one into the other. If (click) doesn't stay an
    option the exerience will be weird).

Issue #140, opened 2016-12-06
Autocomplete components (+html tags?)


MichaelRFairhurst:
When a user starts an html tag

```
  <|
```

we should be able to autocomplete their angular components for them. Probably we should also mix in html tags.

zoechi:
And also tags like `<ng-content>`. Don't remember others provided by Angular that are not components or directives.

kulshekhar:
> Probably we should also mix in html tags

I'm not entirely sure but wouldn't most editors/IDEs do this? Or is it that the angular analyzer would be the sole source of intelligence, overriding everything else?

MichaelRFairhurst:
@kulshekhar I agree, I think that we'll let the IDE do this work for us.

@mk13 can you take this?

You'll want to work off of autocomplete-ast in my fork, and create a PR into it when you're ready.

For inspiration, you can see how I'm handling the case of 

```
  <my-tag |
```

In `server_plugin/lib/src/completion.dart` you'll see I have code to find where a user has their cursor, and if they are in an `ElementInfo` I can suggest `[input]` and `(output)`. Note that my example implementation is incomplete; when a user types in any of the characters for the input, its suddenly no longer targeting an `ElementInfo` but an `ExpressionBoundAttribute` or `StatementBoundAttribute`, and I'm not suggesting anything in that case even though I should. This will come up here: suggestion `<my-tag>` from scratch and/or based on `<|` or `<m|` will be different target elements.

Tests can go in in `server_plugin/test/completion_contributor_test.dart` and the current functionality for suggesting inputs/outputs is in `server_plugin/lib/src/completion.dart` in `TemplateCompleter`, you can add to that or create an abstraction if you want.

We should suggest tag names based on the selectors of the directives that are available. Note that the selectors use an AST, so finding `TagNameSelector`s may require digging through some `AndSelector`s etc. But this should give you a list of things to suggest.


You'll need to use a custom plugin I can email you, and it only works with IJ 2016.3...you can install that via downloading from intellij, and I'll give you instructions on how to keep it from being automatically disabled after you run it.

Any issues you find with like, the completion popup not appearing is either something we need to report to the intelliJ team or a server crash we need to fix. Keep an eye out in general for where IntelliJ does the wrong thing since they have their own release cycle so the earlier we can report it the better.

mk13:
Sure I can take this issue.

mk13:
[...]

PR #202, opened 2016-12-31
Issue140 autocomplete html tags refs #140

mk13:
Html tag completion for defined selectors and exportAs. 

Default HTML tags and default Angular tags are not added yet.

Dependent upon PR: https://github.com/dart-lang/html/pull/46


analyzer_plugin/lib/ast.dart
```diff
@@ -181,7 +181,9 @@ class Mustache extends AngularAstNode {
 /**
  * The HTML elements in the tree
  */
-abstract class NodeInfo extends AngularAstNode {}
+abstract class NodeInfo extends AngularAstNode {
+  int get nodeEnd;
```
MichaelRFairhurst: shouldn't this just be AngularAstNode.offset + AngularAstNode.length? I don't think we need this

analyzer_plugin/lib/ast.dart
```diff
@@ -238,6 +243,11 @@ class ElementInfo extends NodeInfo implements HasDirectives {
   int get length => closingSpan == null
       ? openingSpan.length
       : closingSpan.offset + closingSpan.length - openingSpan.offset;
+  int get nodeEnd =>
+      childNodesMaxEnd ??
+      (closingSpan == null ? null : closingSpan.offset + closingSpan.length) ??
```
MichaelRFairhurst: Awesome usage of ??. Still, this should be equivalent to offset + length, any more complexity than that means either we've double coded it or that our offset or length is returning the wrong info

server_plugin/lib/src/completion.dart
```diff
@@ -27,12 +28,21 @@ bool offsetContained(int offset, int start, int length) {
...
       var target = findTarget(offset, child);
       if (!(target is ElementInfo && target.openingSpan == null)) {
         return target;
       }
+      //Detect unterminated opening html bracket
+    } else if (child is ElementInfo &&
+        child.openingSpan != null &&
+        !offsetContained(offset, child.offset, child.length) &&
+        child.childNodesMaxEnd != null &&
+        offset <= child.childNodesMaxEnd &&
```
MichaelRFairhurst: checking that the `offset <= child.childNodesMaxEnd` should be handled by `offsetContained(offset, length)`, not be custom logic. 

Which makes me wonder, what are the checks for `child.openingSpan != null` and `child.children.isNotEmpty` doing? In the former case, it should be unnecessary since its in the else if, and in the second case, if children is empty it should still be possible for it to be the target of an autocomplete, no?

server_plugin/lib/src/completion.dart
```diff
@@ -249,6 +277,38 @@ class TemplateCompleter {
         location: location, returnType: type.toString());
   }
 
+  CompletionSuggestion _createHtmlTagSuggestion(
+      Component component, int defaultRelevance, protocol.Element element) {
+    String completion = ((component.exportAs != null)
```
MichaelRFairhurst: exportAs actually doesn't determine the html tagname, its what makes `#form="ngForm"` work

server_plugin/lib/src/completion.dart
```diff
@@ -173,32 +183,50 @@ class TemplateCompleter {
...
       }
     }
 
     return suggestions;
   }
 
+  suggestHtmlTags(Template template, List<CompletionSuggestion> suggestions) {
+    for (AbstractDirective abstractDirective in template.view.directives) {
+      if (abstractDirective is Component) {
+        suggestions.add(_createHtmlTagSuggestion(
```
MichaelRFairhurst: Not all directives/components should have corresponding tags. This might need to check if `_createHtmlTagSuggestion` returns null

server_plugin/lib/src/completion.dart
```diff
@@ -249,6 +277,38 @@ class TemplateCompleter {
         location: location, returnType: type.toString());
   }
 
+  CompletionSuggestion _createHtmlTagSuggestion(
+      Component component, int defaultRelevance, protocol.Element element) {
+    String completion = ((component.exportAs != null)
+        ? component.exportAs.name
+        : component.selector.toString());
```
MichaelRFairhurst: This might be more than just an html tag, ie `my-component:not([multiline])`. The selectors are an AST, and to get tag name suggestions we need to walk the selector AST to find which tagnames could be used. For instance, there may be some selector `[prop], tagname` which should match `<tagname>` even though the selector is a tree that starts with OR.

server_plugin/test/completion_contributor_test.dart
```diff
@@ -582,4 +717,203 @@ class OtherComp {
...
+        '/completionTest.dart',
+        '''
+      import '/angular2/angular2.dart';
+      @Component(templateUrl: 'completionTest.html', selector: 'a',
+        directives: const [MyChildComponent])
+        class MyComp{}
+      @Component(template: '', selector: 'my-child-component', exportAs:'myChildComponentAlias')
+      class MyChildComponent{}
+      ''');
+    addTestSource('''<div><div></div></div><^''');
```
MichaelRFairhurst: should be able to autocomplete tags without hitting `<`, based on context where we know a tag could be reasonably placed.

server_plugin/test/completion_contributor_test.dart
```diff
@@ -135,6 +135,141 @@ class MyComp {
     assertSuggestGetter('text', 'String');
     assertSuggestGetter('description', 'String');
   }
+
+  test_completeInlineHtmlSelectorTag_at_beginning() async {
+    addTestSource('''
+import '/angular2/angular2.dart';
+@Component(template: '<^<div></div>', selector: 'my-parent', directives: const[MyChildComponent])
+class MyParentComponent{}
+@Component(template: '', selector: 'my-child')
```
MichaelRFairhurst: should have tests for more complex component selectors

server_plugin/lib/src/completion.dart
```diff
@@ -249,6 +277,38 @@ class TemplateCompleter {
...
+  protocol.Element _createHtmlTagElement(
+      Component component, protocol.ElementKind kind) {
+    String name;
+    int offset, length;
+    if (component.exportAs != null) {
+      name = component.exportAs.name;
+      offset = component.exportAs.nameOffset;
+      length = component.exportAs.nameLength;
+    } else {
+      AngularElement nameElement =
```
MichaelRFairhurst: these won't always be ElementNameSelectors, unfortunately

server_plugin/lib/src/completion.dart
mk13: Sure, will modify this and respective test cases.

analyzer_plugin/lib/ast.dart
mk13: The naming might not be the most accurate, but this variable should be different than normal offset + length. This value stores the absolute offset end of the current ElementInfo node along with its children. In the case that the node is a TextInfo, it is always offset + length. But for ElementInfo, it can potentially be a parent or purely a leaf. I'll explain this purpose in a different comment block.

server_plugin/lib/src/completion.dart
mk13: 'offsetContained' only checks if the direct node contains the provided offset. The recursion scanning handled by 'findTarget' typically works fine, but during unclosed html tags, the DOM becomes mangled making the 'offsetContained' boolean value unreliable. 

The 'child.openingSpan != null' and 'child.children.isNotEmpty' is actually not necessary; just some leftover that I forgot to remove. 

I'll see if there's a way to integrate this logic into 'offsetContained'.

server_plugin/lib/src/completion.dart
mk13: Good catch. Will add the change.

mk13:
Additions needed: 

Handle more complicated cases such as : my-component:not([multiline]) and tests for those. 

Remove 'exportAs'.

Handle case where tag is not always an ElementNameSelector.

Autocomplete on location without needing '<'

server_plugin/lib/src/completion.dart
```diff
@@ -173,32 +183,50 @@ class TemplateCompleter {
...
+        } else {
+          suggestHtmlTags(template, suggestions);
         }
       } else if (target is ExpressionBoundAttribute &&
           target.bound == ExpressionBoundType.input &&
           offsetContained(request.offset, target.originalNameOffset,
               target.originalName.length)) {
         suggestInputs(target.parent.directives, suggestions);
+      } else if (target is TextInfo &&
+          identical(target.text.trimLeft()[0], "<")) {
```
MichaelRFairhurst: This won't work for `<div>some text <^`

mk13:
Revamped the system to calculate html tags while appropriately filtering through non ElementNameSelectors from directives. Calculation now occurs during the same task where directives are processed. 

The element tags are saved within the directive objects. There is some groundwork implemented to later allow for duplicate component-based, tag-name detection and appropriately raise the error (ex: using maps in finding appropriate tag to use). 

Detection now occurs at location rather than through the presence of "<". Based on location, the "<" will also be automatically appended into the autocompletion. The end ">" tag is not suggested since it provides a better work flow; for example, autocomplete will suggest "<someTag" and the user can continue to add classes, attributes, etc rather than being forced to close. 

If this PR is approved, the next PR will be to refactor the completion.dart file since the process is starting to get a bit messy.

analyzer_plugin/lib/ast.dart
```diff
@@ -295,9 +296,11 @@ class ElementInfo extends NodeInfo implements HasDirectives {
       this.templateAttribute);
 
   int get offset => openingSpan.offset;
-  int get length => closingSpan == null
-      ? openingSpan.length
-      : closingSpan.offset + closingSpan.length - openingSpan.offset;
+  int get length => childNodesMaxEnd != null
+      ? childNodesMaxEnd - offset
```
mk13: Oh that would work. My biggest concern was of cases of unclosed open tags, but that should resolve itself through the DOM structure.

analyzer_plugin/lib/src/converter.dart
```diff
@@ -50,7 +52,11 @@ class HtmlTreeConverter {
         attribute.parent = element;
       }
 
-      List<NodeInfo> children = _convertChildren(node);
+      if (element.openingSpan != null) {
```
mk13: Seems like this will still need to be in place to ensure leaf ElementInfo nodes still retain a valid starting `childNodesMaxEnd` value.

analyzer_plugin/lib/src/resolver.dart
```diff
@@ -94,9 +103,14 @@ class DartTemplateResolver {
         continue;
       }
       SourceSpan span = parseError.span;
+      if (parseError.errorCode == 'eof-in-tag-name') {
+        extraNodes.add(new TextInfo(span.start.offset,
```
mk13: I wanted to keep the dangling html tag to be consistent with the case of a single `<` - which is treated as a TextInfo. Either case would work, but the reason was the following: 

`<uncompletedTa^` - The HtmlParser throws an 'eof-found-in-tag-name' error in this case, which leads to this never being propagated into the DOM node. 

`<someTag ^` - This is actually properly parsed by the HTML parser and shows up in the DOM. 

As a result, I'm not sure what would be the benefits of using an ElementInfo unless there's something we're missing. Seeing as how a single space would immediately change it to an ElementInfo through the parser, keeping it as a TextInfo keeps it better consistent with how a single `<` is being treated. 

We can further discuss this if necessary.

analyzer_plugin/lib/src/tasks.dart
```diff
@@ -19,14 +21,38 @@ import 'package:analyzer/src/generated/source.dart';
...
+final ListResultDescriptor<AnalysisError> ANGULAR_HTML_DOCUMENT_ERRORS =
+    new ListResultDescriptor<AnalysisError>(
+        'ANGULAR_HTML_DOCUMENT_ERRORS', AnalysisError.NO_ERRORS);
+
+/**
+ * Additional Nodes generated from errors that need to be processed.
+ */
+final ListResultDescriptor<TextInfo> ANGULAR_HTML_DOCUMENT_EXTRA_NODES =
+    new ListResultDescriptor<TextInfo>(
+        'ANGULAR_HTML_DOCUMENT_EXTRA_NODES', null);
```
mk13: Thanks for the catch!

analyzer_plugin/lib/src/tasks.dart
```diff
@@ -240,7 +370,7 @@ class BuildStandardHtmlComponentsTask extends AnalysisTask {
     SourceFactory sourceFactory = contextTarget.context.sourceFactory;
     Source htmlSource = sourceFactory.forUri(DartSdk.DART_HTML);
     return <String, TaskInput>{
-      UNITS: LIBRARY_SPECIFIC_UNITS.of(htmlSource).toListOf(RESOLVED_UNIT5),
+      UNITS: LIBRARY_SPECIFIC_UNITS.of(htmlSource).toListOf(RESOLVED_UNIT),
```
mk13: Just some remainder of when I was testing out something; reverting back to RESOLVED_UNIT5.

analyzer_plugin/lib/src/tasks.dart
```diff
@@ -1741,5 +1949,25 @@ class _BuildStandardHtmlComponentsVisitor extends RecursiveAstVisitor {
   }
 }
 
+List<AnalysisError> filterParserErrors(
```
mk13: Sure. Changed the other one to be more explicit.

analyzer_plugin/test/tasks_test.dart
```diff
@@ -253,6 +582,183 @@ class DirectiveB {
...
+      {
+        expect(directive.elementTags, hasLength(1));
+        Selector selector = directive.elementTags[0];
+        expect(selector, new isInstanceOf<ElementNameSelector>());
+        expect(selector.toString(), 'dir-b');
+      }
+    }
+  }
+
+  void test_Directive_elementTags_OrSelector() {
```
mk13: I'll keep the Component ones then since they are also technically Directives.

mk13:
I'm going to take another look at how `childNodesMaxEnd` is being handled and focus on that. 
MaterialListComponent and Selector might need some discussion to see what's going on.
For the usage of TextInfo for extra nodes, I'd like to keep it as such unless there's a big reason to switch out of it.

mk13:
Following issues will be handled in a separate refactoring pull request: 

1. Better management of offsets, 'synthetic nodes', and and `findTarget` function. 
2. Refactoring of completion.dart to be more readable.

analyzer_plugin/lib/src/tasks.dart
```diff
@@ -204,23 +204,22 @@ final ListResultDescriptor<View> VIEWS_WITH_HTML_TEMPLATES =
         'ANGULAR_VIEWS_WITH_TEMPLATES', View.EMPTY_LIST);
 
 /**
- * A task that scans contents of a file, producing a set of Dart tokens.
- * Modification of [ParseHtmlTask]
+ * A task that scans contents of a HTML file, producing a set of Dart tokens.
```
MichaelRFairhurst: this still has the error: it doesn't produce a set of dart tokens

Issue #204, opened 2017-01-03
Error when binding contains "<"


MichaelRFairhurst:
Super simple repro:

```
  {{<}}
```

The full scope of  what causes this and whether its common during edit workflows is unclear

stack trace:

```
NoSuchMethodError: The getter 'isDynamic' was called on null.
Receiver: null
Tried calling: isDynamic
#0      Object._noSuchMethod (dart:core-patch/object_patch.dart:44)
#1      Object.noSuchMethod (dart:core-patch/object_patch.dart:47)
#2      ResolverVisitor.visitListLiteral (package:analyzer/src/generated/resolver.dart:6234)
#3      ListLiteralImpl.accept (package:analyzer/src/dart/ast/ast.dart:6907)
#4      ResolverVisitor.visitBinaryExpression (package:analyzer/src/generated/resolver.dart:5547)
#5      BinaryExpressionImpl.accept (package:analyzer/src/dart/ast/ast.dart:1218)
#6      SingleScopeResolver._resolveDartAstNode (package:angular_analyzer_plugin/src/resolver.dart:1175)
#7      SingleScopeResolver._resolveDartExpression (package:angular_analyzer_plugin/src/resolver.dart:1187)
#8      SingleScopeResolver.visitMustache (package:angular_analyzer_plugin/src/resolver.dart:935)
#9      Mustache.accept (package:angular_analyzer_plugin/ast.dart:181)
#10     AngularAstVisitor._visitAllChildren (package:angular_analyzer_plugin/ast.dart:38)
#11     AngularAstVisitor.visitTextInfo (package:angular_analyzer_plugin/ast.dart:31)
#12     TextInfo.accept (package:angular_analyzer_plugin/ast.dart:262)
#13     AngularScopeVisitor.visitElementInScope (package:angular_analyzer_plugin/src/resolver.dart:424)
#14     AngularScopeVisitor.visitElementInfo (package:angular_analyzer_plugin/src/resolver.dart:390)
#15     SingleScopeResolver.visitElementInfo (package:angular_analyzer_plugin/src/resolver.dart:923)
#16     ElementInfo.accept (package:angular_analyzer_plugin/ast.dart:309)
#17     AngularScopeVisitor.visitElementInScope (package:angular_analyzer_plugin/src/resolver.dart:424)
#18     AngularScopeVisitor.visitElementInfo (package:angular_analyzer_plugin/src/resolver.dart:390)
[...]

PR #209, opened 2017-01-04
Fix issue #204 error when '<' is in binding.

MichaelRFairhurst:
Missing resolution step compared to the regular resolution workflow.
Well, technically, we were missingtwo. But one of them: collecting
varibales, is not relevant to us.

This should be a good change for stability in general, though we
probably allow expression types we shouldn't like typed array list
literals that angular won't like.


MichaelRFairhurst:
Ping @scheglov I'd ideally like to get your review on this before merging

Issue #199, opened 2016-12-30
`<option [value]="...">` results in "WARNING: The bound input value does not exist on any directives or on the element


zoechi:
```html
<select [(ngModel)]="language"
        (ngModelChange)="update()"
        [disabled]="readonly" class="form-control">
  <option value=""></option>
  <option *ngFor="let lng of languages.keys"
          [value]="lng">foo
 <!--     ^^^^^^^ -->
  </option>
</select>
```
The `<option>` element does have a `value` property https://developer.mozilla.org/en-US/docs/Web/API/HTMLOptionElement

MichaelRFairhurst:
We scan a dart source file for `document.createElement` so that we can see the tagname for that element. Looks like `dialog`, `option`, `media`, `menuitem`, `mod`, `picture` don't use `document.createElement`, and in fact, don't use their respective tag names at any point in the dart source at all.

To make things worse there's `TableSectionElement` which is an interface, not an element at all, yet implements HTMLElement and isn't abstract. I don't know if there's any programmatic way to tell that this isn't a tag.

I will probably hardcode the acceptance of these tags based on their classname and match the tagname by hand...

MichaelRFairhurst:
Merged #214

PR #214, opened 2017-01-05
Fix issue #199 option value not recognized

MichaelRFairhurst:
Turns out that the option tag, as well as a few others, don't specify
their tagname anywhere in the source of the dartium html file we scan
to generate the standard html components, so it wasn't being recognized
as a component at all.

There's even one component in there that appears at first glance to be
an html component in the same boat, but it isn't. And there's a few
deprecated ones that are like that too. Skip those (testing the
former...the latter...that I care less about).

Just hardcoded a list, and tested that option works just to test that
my list takes effect. Could have tested the rest, but I don't usually
like it when test code and real code match up 1:1 like that would
appear.

Also modify the mock sdk to include the tested cases.

### DIFF
--- a/analyzer_plugin/lib/src/tasks.dart
+++ b/analyzer_plugin/lib/src/tasks.dart
@@ -1540,6 +1540,16 @@ class _BuildStandardHtmlComponentsVisitor extends RecursiveAstVisitor {
   final Map<String, OutputElement> events;
   final Source source;
 
+  static const Map<String, String> specialElementClasses =
+      const <String, String>{
+    "OptionElement": 'option',
+    "DialogElementi": "dialog",
+    "MediaElement": "media",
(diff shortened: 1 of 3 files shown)


analyzer_plugin/lib/src/tasks.dart
```diff
@@ -1540,6 +1540,16 @@ class _BuildStandardHtmlComponentsVisitor extends RecursiveAstVisitor {
   final Map<String, OutputElement> events;
   final Source source;
 
+  static const Map<String, String> specialElementClasses =
+      const <String, String>{
+    "OptionElement": 'option',
+    "DialogElementi": "dialog",
```
mk13: Typo on the end 'i'? Or intentional?

MichaelRFairhurst:
Thanks for the typo find!

Issue #220, opened 2017-01-09
*deferred reports error expected identifier


MichaelRFairhurst:
We have a test case for this, but its not reporting this error for some reason.

MichaelRFairhurst:
Merged #221

PR #221, opened 2017-01-10
Fix issue #220 *deferred error. Change ngFor autocomplete strategy

MichaelRFairhurst:
Previously I had made ngFor="let x of " try to parse a statement after
"of". Unfortunately, this is not actually a valid parsing requirement
(though it does indicate an issue, that's for a linter to raise, or for
us to raise in tandem with special ngFor detection, because in a general
sense it could be valid).

Also noticed that "*structuralDirective="let x; attr; attr;" would be
reported wrong while messing with the idea of treating *deferred as
*deferred=";" to suppress the error. Left that fix in.

Once we don't report "of" itself as a syntactic error (it isn't as said
above), then we no longer have a partial expression for
DartSnippetExtractor to pick up. So instead, inside DartSnippetExtractor
we look for TextAttributes that could be intended to become
ExpressionBoundAttributes, and create an expression to pass to the dart
completer when that happens.

In the process of doing that, found some issues with findTarget in the
presence of stars. Firstly, name, value, and originalName were getting
all messed up, which I fixed. Secondly, the name length isn't actually
based on the length of the name, as in the case of ngForOf which is
seven characters but only appears as "of" in the template with a length
of two. Therefore I made sure that findTarget goes off of originalName,
and that originalName is correct in the virtual attributes.

Now with text attributes having a separate name/original name, all
attributes do (templates for the possible star, inputs and outputs for
the brackets, and text for the sugar in a star). Therefore move the
orgiinalName back into AttributeInfo from the previous base class
BoundAttributeInfo. But keep that around because of the children
property it has.

Autocompleting dart inside ngFor definitely got a bit funky, but I
think it is doable for now.

### DIFF
--- a/server_plugin/lib/src/completion.dart
+++ b/server_plugin/lib/src/completion.dart
@@ -8,10 +8,13 @@ import 'package:analysis_server/src/provisional/completion/dart/completion_dart.
 import 'package:analysis_server/src/services/completion/dart/optype.dart';
 import 'package:analysis_server/src/services/completion/dart/type_member_contributor.dart';
 import 'package:analysis_server/src/services/completion/dart/inherited_reference_contributor.dart';
+import 'package:analyzer/error/error.dart';
+import 'package:analyzer/error/listener.dart';
 import 'package:analyzer/task/dart.dart';
 import 'package:analyzer/task/model.dart';
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/element/type.dart';
+import 'package:angular_analyzer_plugin/src/converter.dart';
 import 'package:angular_analyzer_plugin/src/model.dart';
 import 'package:angular_analyzer_plugin/src/tasks.dart';
 import 'package:angular_analyzer_plugin/ast.dart';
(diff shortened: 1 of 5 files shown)


server_plugin/lib/src/completion.dart
```diff
@@ -78,6 +82,32 @@ class DartSnippetExtractor extends AngularAstVisitor {
       dartSnippet = mustache.expression;
     }
   }
+
+  @override
+  visitTemplateAttr(TemplateAttribute attr) {
+    // if we visit this, we're in a template but after one of its attributes.
+    TextAttribute textAttributeToTreatAsInput;
+    for (AttributeInfo subAttribute in attr.virtualAttributes) {
+      if (subAttribute is TextAttribute &&
```
mk13: Probably fine due to how ngFor is deconstructed, but is there any need to break out of the for loop if the if statement is satisfied?

MichaelRFairhurst: No, we want to grab the last text attribute.

In fact...this is probably why autocompletion got "a bit worky" -- I'm not checking that the last one is a text attr. I just check for the last text attr.

Can fix this

MichaelRFairhurst:
Comments in new commit are important:

Autocomplete in star: if the LATEST node WAS a text attr  …
This also has issues where the original naming data was lost. Track
`*ngFor="let x"` as a var with original name `let x` at original offset
of 'let'.

The parser also gives up when it doesn't see an identifier next. Try to
continue parsing.

Then when you autocomplete at "let ^", we know you're in the middle of
a let expression and we can not suggest you complete nonsense like "let
foo()"

PR #225, opened 2017-01-10
Precompletion refactoring

mk13:
Minor changes prior to refactoring completion.dart


analyzer_plugin/lib/ast.dart
```diff
@@ -293,7 +293,13 @@ class ElementInfo extends NodeInfo implements HasDirectives {
       this.closingNameSpan,
       this.isTemplate,
       this.attributes,
-      this.templateAttribute);
+      this.templateAttribute) {
+    if (!isSynthetic) {
+      childNodesMaxEnd = offset + length;
```
MichaelRFairhurst: Still not sure about this, `offset + length` is the end of _this_ node, not the end of its last child. Thoughts?

analyzer_plugin/lib/src/resolver.dart
```diff
@@ -103,9 +103,20 @@ class DartTemplateResolver {
...
+          parseError.errorCode == 'expected-attribute-name-but-got-eof') {
+        String localName =
+            fragmentText.substring(span.start.offset + "<".length);
+        ElementInfo extraNode = new ElementInfo(
+            localName.trimRight(),
+            new SourceRange(span.start.offset, span.length),
+            null,
+            new SourceRange(span.start.offset + "<".length, localName.length),
+            null,
+            false,
```
MichaelRFairhurst: this should be `localName == "template"` instead of `false`, though I can't think of any fallout from doing it this way

analyzer_plugin/lib/src/tasks.dart
```diff
@@ -267,10 +267,19 @@ class AngularParseHtmlTask extends SourceBasedAnalysisTask {
...
+            parseError.errorCode == 'expected-attribute-name-but-got-eof') {
+          int localNameOffset = span.start.offset + "<".length;
+          String localName = content.substring(localNameOffset).trimRight();
+          ElementInfo extraNode = new ElementInfo(
+              localName.trimRight(),
+              new SourceRange(span.start.offset, span.length),
+              null,
+              new SourceRange(localNameOffset, localName.length),
+              null,
+              false,
```
MichaelRFairhurst: same comment regarding `localName == "template"`

analyzer_plugin/lib/ast.dart
mk13: This is to ensure that the childNodesMaxEnd is initialized for any leaf ElementInfo. The 'childNodesMaxEnd' is propagated upwards in the DOM, so if it's not initialized, the propagation will never work.

Issue #223, opened 2017-01-10
Supply proper completionLength, completionOffset


MichaelRFairhurst:
Looks like IntelliJ will start using these offsets to let us deal with the complex problem of knowing what to replace. Just in general, we should assume that all editors use this info (though IntelliJ used to ignore it and I think vim ignores it too).

zoechi:
Is this related or should I create a new issue? (missing ` != null`)

![image](https://cloud.githubusercontent.com/assets/405837/22248543/53d3c018-e23f-11e6-8877-b89f0593b9b8.png)

MichaelRFairhurst:
this only has to do with autocompletion

zoechi:
Sorry, didn't read properly. I removed the comment and created #241

MichaelRFairhurst:
this is done...merged in PR... #226

PR #226, opened 2017-01-10
Solve issue #223 supply completion replacement length & offset

MichaelRFairhurst:
Added the ideal replacement length & offset assertions in the tests and
went from there.

Create a new visitor to detect what should be replaced in a target, but
also let the dart completion infrastructure override that when
'extractor.dartSnippet' is not null.

Always suggest `<` in tags since we can tell intelliJ to replace those.

Also found an issue where autocompleting `blah="^"` resulted in a
`replacementLength` of 4, forgotten I had treated that as `blah="null"`
in the converter to suppress the "expected identifier" warning since we
had a later warning for the tag being empty already and didn't want to
report two.

So no longer magically insert a null which messes up completion
lengths, but do still suppress that warning using a new class
IgnoringAnalysisErrorListener. Note that I had created a class with the
same name in a different branch to solve #220, when these get merged
those two classes should use the implementation in this branch where
its in its own file.

Not currently working for "<x^<y" because the html parser treats that
as one tag. And not working for "<x" at EOF because that's treated as a
text node. These should be solved by modifying the html parser ideally.

### DIFF
--- a/server_plugin/lib/src/completion.dart
+++ b/server_plugin/lib/src/completion.dart
@@ -129,10 +129,10 @@ class ReplacementRangeCalculator extends AngularAstVisitor {
     if (element.openingSpan == null) {
       return;
     }
-    if (offsetContained(
-        request.offset,
-        element.openingNameSpan.offset - '<'.length,
-        element.openingNameSpan.length + '<'.length)) {
+    int nameSpanEnd =
+        element.openingNameSpan.offset + element.openingNameSpan.length;
+    if (offsetContained(request.offset, element.openingSpan.offset,
+        nameSpanEnd - element.openingSpan.offset)) {
       request.replacementOffset = element.openingSpan.offset;
       request.replacementLength = element.localName.length + 1;
     }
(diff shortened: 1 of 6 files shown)


server_plugin/lib/src/completion.dart
```diff
@@ -116,15 +118,72 @@ class LocalVariablesExtractor extends AngularAstVisitor {
...
+
+  // don't recurse, findTarget already did that
+  @override
+  visitElementInfo(ElementInfo element) {
+    if (element.openingSpan == null) {
+      return;
+    }
+    if (offsetContained(
+        request.offset,
+        element.openingNameSpan.offset - '<'.length,
```
mk13: Can't this just be `element.openingSpan.offset`?

mk13:
Looks good!

server_plugin/lib/src/completion.dart
MichaelRFairhurst: Makes more sense, doesn't it? +1

MichaelRFairhurst:
I improved the length calculation too

MichaelRFairhurst:
@mk13 how does this look from a delivery perspective? We need to fix the `<div<div` problem before all test cases will pass, and we need to use this for intelliJ to work. What's the timeline for unblocking this?

mk13:
We can push the temporary html parser solution early this week and return back to this for incorrect edge cases (ex: <div<div>)

PR #230, opened 2017-01-18
Track, validate, and autocomplete transclusions.

MichaelRFairhurst:
In order to track transclusions, we need to look into the html
templates of child components before we can check their parents. The
trick there is that this can be cyclic.

But its not really that its cyclic, its really just that its a first
pass. We don't need to know the relationship between the views before
we get the <ng-content>s in the html file for the view. Therefore we've
split VIEWS into VIEWS1 and VIEWS2. VIEWS1 is for getting the minimum
data to build an AST to check for <ng-content> selectors. In VIEWS2 we
wire up the Views that all have ASTs and ngContent tags.

We also have ANGULAR_ASTS as a new target which takes a dart file and
builds the ASTs for each view in that dart file. It is where we parse
inline dart templates, allowing us to consolidate html parsing into a
mixin in src/tasks.dart. It does, however, produce errors for the
templates not the dart file. So that's an output which is lists of
errros mapped to sources.

Then when you resolve a dart or html template, you make sure to depend
upon ANGULAR_ASTS.of(included directives dart files).

New visitor to collect the <ng-content>s, since it requires no
resolution and can be done in a single pass.

In the future, some changes could be made so that when analyzing a
component template, only the child directive & its <ng-content>s are
needed, but the grandchilds aren't (by depending on VIEWS1).

Also create an API for Selectors to produce HTML tag suggestions since
transclusions can be very hard to figure out. Autocompleting them, even
if djanky, is good. Use a constraining/copying system.

Now in order to suggest the proper elements, we need to track the
parents of TextInfos (and ElementInfos too for less obvious reasons).
Track those in converter.dart.

Also, when validating translusions, if a tag is matched because of its
tagname, record that so that the directive resolver doesn't flag it as
unknown.

We still need to handle @ContentChild, I won't know if that comes up
internally until I try this out later.

### DIFF
--- a/analyzer_plugin/lib/src/selector.dart
+++ b/analyzer_plugin/lib/src/selector.dart
@@ -1,5 +1,7 @@
 library angular2.src.analysis.analyzer_plugin.src.selector;
 
+import 'dart:collection';
+
 import 'package:analyzer/src/generated/source.dart';
 import 'package:angular_analyzer_plugin/src/model.dart';
 import 'package:angular_analyzer_plugin/src/strings.dart';
@@ -7,7 +9,7 @@ import 'package:angular_analyzer_plugin/src/strings.dart';
 /**
  * The [Selector] that matches all of the given [selectors].
  */
-class AndSelector implements Selector {
+class AndSelector extends Selector {
   final List<Selector> selectors;
 
   AndSelector(this.selectors);
@@ -27,13 +29,21 @@ class AndSelector implements Selector {
 
   @override
   String toString() => selectors.join(' && ');
+
+  List<HtmlTagForSelector> refineTagSuggestions(
+      List<HtmlTagForSelector> context) {
+    for (Selector selector in selectors) {
+      context = selector.refineTagSuggestions(context);
+    }
+    return context;
+  }
 }
 
 /**
  * The [Selector] that matches elements that have an attribute with the
  * given name, and (optionally) with the given value;
  */
-class AttributeSelector implements Selector {
+class AttributeSelector extends Selector {
   final AngularElement nameElement;
   final bool isWildcard;
   final String value;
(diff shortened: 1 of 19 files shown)


analyzer_plugin/lib/src/selector.dart
```diff
@@ -298,6 +378,98 @@ class SelectorParseError extends FormatException {
...
+  bool _isValid = true;
+  Set<String> _classes = new HashSet<String>();
+
+  bool get isValid => _name != null && _isValid && _classAttrValid;
+
+  bool get _classAttrValid => _classes.isEmpty || _attributes["class"] == null
+      ? true
+      : _classes.length == 1 && _classes.first == _attributes["class"];
+
+  void set name(String name) {
```
mk13: A bit confused here; does this make the name either 'null and mutable once OR immutable once assigned'? If so, what's the downside to making the field final and just initializing it upon construction?

MichaelRFairhurst: I'm trying to track invalid selectors, such as `foo:not(foo)` (although I'm not actually tracking that one....)

In a more complex example, `([foo], [bar])([whiz], [bang])` matches `<_ foo whiz`, `<_ foo bang`, `<_ bar whiz`, and `<_ bar bang`. But if you make "foo" and "whiz" into element names, ie, "(foo, [bar])(whiz, [bang])` then only `<foo bang` and `<whiz bar` are valid. Still, sometimes that's the best way to express a boolean expression.

Basically each selector constrains the set of elements that match, and I'm trying to detect when the resulting constraints have no overlap, and then set isValid = false. The resulting state otherwise can get mangled, because isValid is false already.

Looks though like only real life case I caught was stuff like "[foo=bar][foo=bang]", which is garbage anyway, but handling ":not" is much much harder than handling conflicting tagnames, so I did what was easy.

PR #243, opened 2017-01-25
Temporary AngularHtmlParser implemented to proceed onto offsets for autocompletion

mk13:
Temporary AngularHtmlParser added to create synthetic nodes. This is planned to be replaced by `angular_ast` upon completion.


analyzer_plugin/lib/src/converter.dart
```diff
@@ -130,6 +143,14 @@ class HtmlTreeConverter {
...
   }
 
+  TemplateAttribute _convertSyntheticAttribute(html.Element element) {
+    FileSpan openSourceSpan = element.sourceSpan;
+    int origNameOffset = openSourceSpan.start.offset + openSourceSpan.length;
+    TemplateAttribute templateAttribute = new TemplateAttribute(
+        "", origNameOffset, null, null, "", origNameOffset, []);
+    return templateAttribute;
+  }
+
```
MichaelRFairhurst: Why is this creating a template attribute? That's a `*x=` or `template=` attr

mk13: Meant to use TextAttribute. Fixed.

PR #247, opened 2017-02-15
resolve #229 more ngFor special variables (odd, even, first, last)

### LINKED ISSUE #229: expose more ngFor special variables
Variables include even, odd, I think first, last too...maybe more things.

MichaelRFairhurst:


zoechi:
Every directive or component that stamps a template with `createEmbeddedView` (perhaps others) can pass all kinds of variables as context. Shouldn't the available variables not just be derived from the `<template let-x>` "declarations" on the template element to make it more universal? I guess there is currently no way to get the type though.

PR #270, opened 2017-04-11
Track @Attribute annotated constructor parameters

MichaelRFairhurst:
We can then autocomplete these, and one step closer to being able to flag unrecognized Attributes.

Also, didn't know these existed until yesterday....


analyzer_plugin/lib/src/directive_extraction.dart
```diff
@@ -443,6 +443,44 @@ class DirectiveExtractor extends AnnotationProcessorMixin {
...
+            // TODO do we ever need to report an error here, or will DAS?
+          }
+
+          if (parameter.type.name != "String") {
+            errorReporter.reportErrorForOffset(
+                AngularWarningCode.ATTRIBUTE_PARAMETER_MUST_BE_STRING,
+                parameter.nameOffset,
+                parameter.name.length);
+          }
+
```
mk13: Just wondering: if the declared type is not a String, do we still want to store it in attributes list? 

Are we keeping this variable and treating it as a String type while just throwing an error so the user knows to change from non-String to String?

MichaelRFairhurst: Yeah, if someone uses @Attribute("foo") int foo, we can still suggest "foo" for the usage of the tag. But that's about all that we'll ever use this info for

PR #273, opened 2017-04-11
Now ready: Priority angular analysis

MichaelRFairhurst:
Depends on: https://codereview.chromium.org/2815643004/

Turns out this is also required for vim.

Still have to improve requested html analysis, as it currently just assumes `x.dart` and `x.html` relate. This is because otherwise we'd have to wait until allllll dart analyses finish before we can start the priority html analysis. However, sometimes all dart is already analyzed. In this case, we should go off of what we know are the latest links between files.


MichaelRFairhurst:
Ping @mk13 for re-review of latest changes

analyzer_plugin/lib/src/angular_driver.dart
```diff
@@ -189,7 +179,19 @@ class AngularDriver
       // driver, which doesn't happen because its waiting for us.
       // ALSO assume .dart and .html paths correlate, otherwise we'd have to
       // wait for all dart analysis to complete.
-      resolveHtml(path, path.replaceAll(".html", ".dart")).then((result) {
+      Future resolvedHtml;
+
+      // Try resolving HTML using the existing dart/html relationships which may
+      // be already known. However, if we don't see any relationships, try using
+      // the .dart equivalent. Better than no result -- the real one WILL come.
+      if (_fileTracker.getDartPathsReferencingHtml(path) == 0) {
```
mk13: Doesn't this function return a List? Did you mean .length?

MichaelRFairhurst: Yes, great catch!

PR #276, opened 2017-04-11
Summary integration

mk13:
One more try.


analyzer_plugin/lib/src/converter.dart
```diff
@@ -7,52 +7,183 @@ import 'package:analyzer/src/dart/ast/token.dart' hide SimpleToken;
...
 import 'package:angular_analyzer_plugin/src/ng_expr_parser.dart';
 import 'package:angular_analyzer_plugin/src/ignoring_error_listener.dart';
-import 'package:angular_analyzer_plugin/src/angular_html_parser.dart';
 import 'package:angular_analyzer_plugin/src/strings.dart';
 import 'package:angular_analyzer_plugin/tasks.dart';
-import 'package:html/dom.dart' as html;
-import 'package:html/parser.dart' as html;
-import 'package:source_span/source_span.dart';
-
-html.Element firstElement(html.Node node) {
```
MichaelRFairhurst: Nice we can remove this :)

analyzer_plugin/lib/src/converter.dart
```diff
@@ -7,52 +7,183 @@ import 'package:analyzer/src/dart/ast/token.dart' hide SimpleToken;
...
-}
 
 class HtmlTreeConverter {
   final EmbeddedDartParser dartParser;
   final Source templateSource;
   final AnalysisErrorListener errorListener;
 
+  // Following Angular2 Logic:
+  // https://github.com/dart-lang/angular2/blob/8220ba3a693aff51eed33cd1ec9542bde9017423/lib/src/compiler/schema/dom_element_schema_registry.dart#L199
+  static const attrToPropMap = const {
```
MichaelRFairhurst: should this be done in angular_ast?

analyzer_plugin/lib/src/converter.dart
```diff
@@ -7,52 +7,183 @@ import 'package:analyzer/src/dart/ast/token.dart' hide SimpleToken;
...
+    for (StandaloneTemplateAst node in asts) {
+      var convertedNode = convert(node, parent: root);
+      if (convertedNode != null) {
+        root.childNodes.add(convertedNode);
+      }
+    }
+    return root;
+  }
+
+  NodeInfo convert(StandaloneTemplateAst node, {ElementInfo parent}) {
```
MichaelRFairhurst: parent is no longer optional, really, right?

You could make it just a second parameter, or to keep the `parent:` naming (kinda nice right?) you can mark it `@required` which will make it an error if we forget to use `parent:` even though its a named (as opposed to positional) param.

analyzer_plugin/lib/src/converter.dart
```diff
@@ -7,52 +7,183 @@ import 'package:analyzer/src/dart/ast/token.dart' hide SimpleToken;
...
+      SourceRange closingSpan;
+      SourceRange closingNameSpan;
+
+      if (node.isSynthetic) {
+        openingSpan = _toSourceRange(closeComponent.beginToken.offset, 0);
+        openingNameSpan = openingSpan;
+      } else {
+        openingSpan = _toSourceRange(
+            node.beginToken.offset, node.endToken.end - node.beginToken.offset);
+        openingNameSpan =
```
MichaelRFairhurst: there's no real opening/closing name span in angular ast? Would be nice to not make it `openingNameSpan - '<'.length`, but something tracked at parse time.

analyzer_plugin/lib/src/converter.dart
```diff
@@ -7,52 +7,183 @@ import 'package:analyzer/src/dart/ast/token.dart' hide SimpleToken;
...
+
+      if (node.isSynthetic) {
+        openingSpan = _toSourceRange(closeComplement.beginToken.offset, 0);
+        openingNameSpan = openingSpan;
+      } else {
+        openingSpan = _toSourceRange(
+            node.beginToken.offset, node.endToken.end - node.beginToken.offset);
+        openingNameSpan =
+            new SourceRange(openingSpan.offset + '<'.length, localName.length);
+        var pnode = node as ParsedEmbeddedContentAst;
```
MichaelRFairhurst: Why is this cast necessary? Can't the Parsed and Synthetic versions of EmbeddedContentAst both share the same interface?

analyzer_plugin/lib/src/converter.dart
```diff
@@ -75,98 +206,213 @@ class HtmlTreeConverter {
...
+    if (node is TextAst) {
       int offset = node.sourceSpan.start.offset;
-      String text = node.text;
+      String text = node.value;
+      return new TextInfo(
+          offset, text, parent, dartParser.findMustaches(text, offset));
+    }
+    if (node is InterpolationAst) {
+      int offset = node.sourceSpan.start.offset;
+      String text = '{{' + node.value + '}}';
```
MichaelRFairhurst: Is this accurate when there are spaces? Does `node.value` include them?

analyzer_plugin/lib/src/converter.dart
```diff
@@ -75,98 +206,213 @@ class HtmlTreeConverter {
...
-            attributes.add(new TextAttribute(
-                name,
-                _nameOffset(element, name),
-                value,
-                valueOffset,
-                dartParser.findMustaches(value, valueOffset)));
+
+    // Atttribute/event/properties/etc. within
+    // [ElementAst] cannot be synthetic as long as Desugaring never occurs.
+    if (element is ElementAst) {
```
MichaelRFairhurst: unnecessary check

analyzer_plugin/lib/src/converter.dart
```diff
@@ -75,98 +206,213 @@ class HtmlTreeConverter {
...
+      for (AttributeAst attribute in element.attributes) {
+        if (attribute.name.startsWith('on-')) {
+          attributes
+              .add(_convertStatementsBoundAttribute(attribute, "on-", null));
+        } else if (attribute.name.startsWith('bind-')) {
+          attributes.add(_convertExpressionBoundAttribute(
+              attribute, "bind-", null, ExpressionBoundType.input));
```
MichaelRFairhurst: Shouldn't these be handled in angular_ast?

analyzer_plugin/lib/src/converter.dart
```diff
@@ -75,98 +206,213 @@ class HtmlTreeConverter {
...
+        } else if (attribute.name == 'template') {
+          attributes.add(_convertTemplateAttribute(attribute));
```
MichaelRFairhurst: same

analyzer_plugin/lib/src/converter.dart
```diff
@@ -75,98 +206,213 @@ class HtmlTreeConverter {
...
-    TextAttribute textAttribute =
-        new TextAttribute("", nameOffset, null, null, []);
-    return textAttribute;
+  List<AttributeInfo> _convertAttributesForTemplate(
+      EmbeddedTemplateAst element) {
+    List<AttributeInfo> attributes = <AttributeInfo>[];
+
+    // Atttribute/event/properties/etc. within
+    // [ElementAst] cannot be synthetic as long as Desugaring never occurs.
+    if (element is EmbeddedTemplateAst) {
```
MichaelRFairhurst: also unnecessary

analyzer_plugin/lib/src/converter.dart
```diff
@@ -75,98 +206,213 @@ class HtmlTreeConverter {
...
+            value = _attr.valueToken.innerValue.lexeme;
+            valueOffset = _attr.valueToken.innerValue.offset;
+          }
+          attributes.add(new TextAttribute(_attr.name, _attr.nameOffset, value,
+              valueOffset, dartParser.findMustaches(value, valueOffset)));
+        }
+      }
+      ;
+
+      for (PropertyAst property in element.properties) {
```
MichaelRFairhurst: Seems like this code is duplicated and could be abstracted. All property asts, reference asts, etc, all convert the same regardless of whether they're in an EmbeddedTemplateAst or ElementAst or not, right?

analyzer_plugin/lib/src/converter.dart
```diff
@@ -75,98 +206,213 @@ class HtmlTreeConverter {
...
+      origName = ast.prefixToken.lexeme + ast.nameToken.lexeme;
+      origNameOffset = ast.prefixToken.offset;
+
+      name = ast.nameToken.lexeme;
+      nameOffset = ast.nameToken.offset;
+
+      String fullAstName;
+      if (value != null) {
+        fullAstName = ast.name +
+            (' ' * (ast.valueToken.innerValue.offset - ast.nameToken.end)) +
```
MichaelRFairhurst: nice. Way cleaner than the `'="'.length` thing. That always bothered me!

analyzer_plugin/lib/src/converter.dart
```diff
@@ -208,24 +492,84 @@ class HtmlTreeConverter {
...
+      origNameOffset = ast.nameOffset;
+
+      value = ast.value;
+      if (value == null || value.isEmpty) {
+        errorListener.onError(new AnalysisError(templateSource, origNameOffset,
+            origName.length, AngularWarningCode.EMPTY_BINDING, [origName]));
+      }
+      valueOffset = ast.valueOffset;
+
+      propName = _removePrefixSuffix(origName, prefix, suffix);
```
MichaelRFairhurst: Looks like this code also has some duplication that could be cleaned up. Like this line, the next line, and checking/reporting empty bindings. I bet these checks could purely turn attributes into variables and all the logic could be afterwards.

Alternatively, any data that you need off of each of the three types could be methods on the AST. "hasSynthetic", "name", "nameOffset", "value", on some interface shared by all properties.

analyzer_plugin/lib/src/converter.dart
```diff
@@ -455,19 +766,24 @@ class EmbeddedDartParser {
...
+        // Both exists, but there is an end before a begin.
+        // Example: blah }} {{ mustache ...
         errorListener.onError(new AnalysisError(templateSource,
             fileOffset + end, 2, AngularWarningCode.UNOPENED_MUSTACHE));
         // Move the cursor ahead and keep looking for more unmatched mustaches.
         textOffset = end + 2;
         continue;
       } else if (nextBegin != -1 && nextBegin < end) {
+        // Two open mustaches, but both opens are in sequence before an end.
+        // Example: {{ blah {{ mustache }}
```
MichaelRFairhurst: Fantastic comments!

analyzer_plugin/lib/src/view_extraction.dart
```diff
@@ -197,55 +189,64 @@ class ViewExtractor extends AnnotationProcessorMixin {
...
 class TemplateParser {
-  html.Document document;
+  static const errorMap = const {
+    NgAst.NgParserWarningCode.UNTERMINATED_MUSTACHE:
+        AngularWarningCode.UNTERMINATED_MUSTACHE,
+    NgAst.NgParserWarningCode.UNOPENED_MUSTACHE:
+        AngularWarningCode.UNOPENED_MUSTACHE,
+  };
+
+  List<NgAst.TemplateAst> document;
```
MichaelRFairhurst: I'm not sure document is the best name for this any longer...

analyzer_plugin/test/resolver_test.dart
```diff
@@ -1015,8 +1017,13 @@ class TestPanel {
 """;
     _addHtmlSource(code);
     await _resolveSingleTemplate(dartSource);
-    assertErrorInCodeAtPosition(
-        AngularWarningCode.INVALID_CSS_PROPERTY_NAME, code, "invalid*property");
+    assertMultipleErrorsExplicit([
+      new AnalysisError(
+          htmlSource, 29, 0, AngularWarningCode.NONEXIST_INPUT_BOUND, ['']),
```
MichaelRFairhurst: Hmm, not a big fan of the fixed offsets. Can we get positions by some friendlier means?

analyzer_plugin/test/resolver_test.dart
```diff
@@ -2249,29 +2272,30 @@ class TestPanel {
...
-</li>
-<li *ngFor="let item3 of items; let i=index" {{lowerEl}}>
-  {{item3.length}}
-</li>
-<div #lowerEl></div>
-""");
-    await _resolveSingleTemplate(dartSource);
-    errorListener.assertNoErrors();
-  }
+  //TODO: Max: Figure out if this is deprecated; if not, what is it?
```
MichaelRFairhurst: Should be deleted

server_plugin/lib/src/completion.dart
```diff
@@ -239,88 +247,83 @@ class TemplateCompleter {
...
 
         dartRequest.libraryElement = template.view.classElement.library;
-        TypeMemberContributor memberContributor = new TypeMemberContributor();
-        InheritedReferenceContributor inheritedContributor =
-            new InheritedReferenceContributor();
-        suggestions.addAll(inheritedContributor.computeSuggestionsForClass(
-            template.view.classElement, dartRequest,
-            skipChildClass: false));
+        var memberContributor = new TypeMemberContributor();
+        var inheritedContributor = new InheritedReferenceContributor();
```
MichaelRFairhurst: Liking the reduction of types that don't help readability. Very nice! Though I might have preferred final, not everyone does, so it can be a controversy worth avoiding :) So looks great!

server_plugin/lib/src/completion.dart
```diff
@@ -239,88 +247,83 @@ class TemplateCompleter {
...
+
+        suggestions.addAll(
+          inheritedContributor.computeSuggestionsForClass(
+            template.view.classElement,
+            dartRequest,
+            skipChildClass: false,
```
MichaelRFairhurst: dangling comma?

server_plugin/lib/src/completion.dart
```diff
@@ -239,88 +247,83 @@ class TemplateCompleter {
...
         if (dartRequest.opType.includeIdentifiers) {
-          LocalVariablesExtractor varExtractor = new LocalVariablesExtractor();
+          var varExtractor = new LocalVariablesExtractor();
           target.accept(varExtractor);
           if (varExtractor.variables != null) {
             addLocalVariables(
-                suggestions, varExtractor.variables, dartRequest.opType);
+              suggestions,
+              varExtractor.variables,
+              dartRequest.opType,
```
MichaelRFairhurst: same?

server_plugin/lib/src/completion.dart
```diff
@@ -239,88 +247,83 @@ class TemplateCompleter {
...
+                target.closingSpan.length)) {
+          // In closing tag, but could be directly after it; ex: '</div>^'.
+          if (request.offset ==
+              (target.closingSpan.offset + target.closingSpan.length)) {
+            suggestHtmlTags(template, suggestions);
+            if (target.parent != null || target.parent is! DocumentInfo) {
+              suggestTransclusions(target.parent, suggestions);
+            }
+          }
+          // Directly within closing tag; suggest nothing. Ex: '</div^>'
```
MichaelRFairhurst: Once again, nice comments.

server_plugin/lib/src/completion.dart
```diff
@@ -239,88 +247,83 @@ class TemplateCompleter {
...
+          else
```
MichaelRFairhurst: brackets

analyzer_plugin/lib/src/converter.dart
mk13: One problem I see with this is that angular_ast doesn't keep track of two different names, which could make the offset off for 'class':'className' conversion. ExpressionBoundAttribute class keeps track of two different names: origName and propName, which allows us some flexibility in what is being interpreted for value and one for offset.

analyzer_plugin/lib/src/converter.dart
mk13: I can use the identifierToken actually so it will be similar to how openingSpan is defined.
openingNameSpan = _toSourceRange(identifierToken.offset, identifierToken.length)

analyzer_plugin/lib/src/converter.dart
mk13: Parsed version have access to tokens, synthetic does not and the tokens contain the direct offsets. As a result, the casting is necessary.

analyzer_plugin/lib/src/converter.dart
mk13: The node.value contains the actual whitespaces, so it should be still be accurate.

analyzer_plugin/lib/src/converter.dart
mk13: I'll look into migrating this logic into the AST; I believe it was raised in one of the issue within that project. If so, I can probably also move the 'attrToPropMap' raised earlier simultaneously.

analyzer_plugin/lib/src/converter.dart
mk13: Will work on this 👍

analyzer_plugin/lib/src/converter.dart
mk13: That's true, I'll look towards condensing this logic into a cleaner abstracted helper function.

server_plugin/lib/src/completion.dart
mk13: Ah turns out if you put a dangling comma at the end and call 'dartfmt', it makes long lists
very neatly aligned. If it's a long list, I think this adds to readability in exchange for some additional lines.

analyzer_plugin/lib/src/converter.dart
MichaelRFairhurst: From our conversation -- agreed. This is the right place for this!

analyzer_plugin/lib/src/converter.dart
MichaelRFairhurst: Interesting. Ok, just wanted to ask!

MichaelRFairhurst: In fact that's great, as compared to getting null unexpectedly.

analyzer_plugin/lib/src/converter.dart
MichaelRFairhurst: ah ok nice

server_plugin/lib/src/completion.dart
MichaelRFairhurst: k, sgtm

analyzer_plugin/lib/src/converter.dart
```diff
@@ -308,64 +311,62 @@ class HtmlTreeConverter {
...
-            value,
-            valueOffset,
-            dartParser.findMustaches(value, valueOffset)));
-      }
-      ;
+        attributes.add(new TextAttribute(_attr.name, _attr.nameOffset, value,
+            valueOffset, dartParser.findMustaches(value, valueOffset)));
+      }
+    }
+    ;
```
MichaelRFairhurst: dangling semi

analyzer_plugin/lib/src/converter.dart
```diff
@@ -308,64 +311,62 @@ class HtmlTreeConverter {
...
+            property, "[attr.", "]", ExpressionBoundType.attr));
+      } else if (property.name.startsWith("style")) {
+        attributes.add(_convertExpressionBoundAttribute(
+            property, "[style.", "]", ExpressionBoundType.style));
+      } else {
+        attributes.add(_convertExpressionBoundAttribute(
+            property, "[", "]", ExpressionBoundType.input));
+      }
+    }
+    ;
```
MichaelRFairhurst: here too

analyzer_plugin/lib/src/converter.dart
```diff
@@ -226,79 +232,76 @@ class HtmlTreeConverter {
...
-      ;
-
-      element.stars.forEach((star) {
-        attributes.add(_convertTemplateAttribute(star));
-      });
+        attributes.add(new TextAttribute(_attr.name, _attr.nameOffset, value,
+            valueOffset, dartParser.findMustaches(value, valueOffset)));
+      }
     }
+    ;
```
MichaelRFairhurst: dangling semi

analyzer_plugin/lib/src/converter.dart
```diff
@@ -226,79 +232,76 @@ class HtmlTreeConverter {
...
+      } else if (property.name.startsWith("style") &&
+          property.postfix != null) {
+        attributes.add(_convertExpressionBoundAttribute(
+            property, "[style.", "]", ExpressionBoundType.style));
+      } else {
+        attributes.add(_convertExpressionBoundAttribute(
+            property, "[", "]", ExpressionBoundType.input));
+      }
+    }
+    ;
```
MichaelRFairhurst: here too

analyzer_plugin/lib/src/converter.dart
```diff
@@ -226,79 +232,76 @@ class HtmlTreeConverter {
...
+        valueOffset = _attr.valueToken.innerValue.offset;
+      }
+      attributes.add(new TextAttribute(
+          _attr.prefixToken.lexeme + _attr.nameToken.lexeme,
+          _attr.prefixToken.offset,
+          value,
+          valueOffset,
+          dartParser.findMustaches(value, valueOffset)));
+    }
+    ;
```
MichaelRFairhurst: here too

analyzer_plugin/lib/src/converter.dart
```diff
@@ -308,64 +311,62 @@ class HtmlTreeConverter {
...
+        valueOffset = _attr.valueToken.innerValue.offset;
+      }
+      attributes.add(new TextAttribute(
+          _attr.prefixToken.lexeme + _attr.nameToken.lexeme,
+          _attr.prefixToken.offset,
+          value,
+          valueOffset,
+          dartParser.findMustaches(value, valueOffset)));
     }
+    ;
```
MichaelRFairhurst: and here

analyzer_plugin/lib/src/view_extraction.dart
```diff
@@ -196,15 +196,15 @@ class TemplateParser {
         AngularWarningCode.UNOPENED_MUSTACHE,
   };
 
-  List<NgAst.TemplateAst> document;
+  List<NgAst.TemplateAst> rawAst;
```
MichaelRFairhurst: great name

analyzer_plugin/test/abstract_angular.dart
```diff
@@ -314,20 +315,36 @@ class NgFor {
...
+   *    code segment where offset begins,
+   *    length of the error highlight,
+   *    errorCode,
+   *    and optional error args - pass empty list if not needed.
    */
-  void assertMultipleErrorsExplicit(List<AnalysisError> expectedErrors) {
+  void assertMultipleErrorsExplicit(
+    Source source,
+    String code,
+    List<Tuple4<String, int, ErrorCode, List<Object>>> expectedErrors,
```
MichaelRFairhurst: nice

analyzer_plugin/test/abstract_angular.dart
```diff
@@ -314,20 +315,36 @@ class NgFor {
...
   }
 
-  /**
-   * Given an explicit list of [AnalysisError], check to see if errors
-   * occurred during angular analysis.
+  /** For [expectedErrors], it is a List of Tuple4 (1 per error):
+   *    code segment where offset begins,
+   *    length of the error highlight,
+   *    errorCode,
+   *    and optional error args - pass empty list if not needed.
```
MichaelRFairhurst: only just realized that your new assertion function here checks the args. I've wanted to have tests for them, this is great!

analyzer_plugin/test/resolver_test.dart
```diff
@@ -1017,12 +1017,11 @@ class TestPanel {
...
     _addHtmlSource(code);
     await _resolveSingleTemplate(dartSource);
-    assertMultipleErrorsExplicit([
-      new AnalysisError(
-          htmlSource, 29, 0, AngularWarningCode.NONEXIST_INPUT_BOUND, ['']),
-      new AnalysisError(htmlSource, 29, 1,
-          NgParserWarningCode.EXPECTED_WHITESPACE_BEFORE_NEW_DECORATOR),
-      new AnalysisError(htmlSource, 6, 14, NgParserWarningCode.SUFFIX_PROPERTY),
+    assertMultipleErrorsExplicit(htmlSource, code, [
+      new Tuple4(']', 0, AngularWarningCode.NONEXIST_INPUT_BOUND, ['']),
```
MichaelRFairhurst: Hmm, hadn't noticed before that these errors had a length of zero. How does intelliJ handle that?

analyzer_plugin/test/abstract_angular.dart
```diff
@@ -314,20 +315,36 @@ class NgFor {
...
-    expectedErrors.forEach((expectedError) {
-      expect(realErrors.contains(expectedError), true,
-          reason:
-              'Expected error code ${expectedError.errorCode} never occurs at '
-              'location ${expectedError.offset} of length ${expectedError.length}.');
-    });
-    expect(realErrors.length, expectedErrors.length,
-        reason: 'Expected error counts do not  match.');
+    for (Tuple4 expectedError in expectedErrors) {
+      var offset = code.indexOf(expectedError.item1);
```
MichaelRFairhurst: Might want to assert that this isn't -1, or maybe even assert that the expected code doesn't occur more than once in the test file?

server_plugin/lib/src/completion.dart
MichaelRFairhurst: Can you maybe move this comment inside the else block? Otherwise I read this as the end of the control structure, takes me a moment to realize that the "else" goes here.

analyzer_plugin/test/resolver_test.dart
mk13: When you double click it it puts the cursor right at the spot, but doesn't squiggle anything since it's of length 0. I'm planning sometime later once the ast is merged to do another pass and make the error messages more helpful in terms of length.

mk13:
Still need to add change such that conversion does not have three semi-redundant conversions based on check of ElementAst, EmbeddedTemplateAst, and EmbeddedContentAst. 

Need to add some refactor into angular_ast to create a generic encapsulation interface and want to re-factor EmbeddedContentAst generation to recycle some code instead of creating its own (seemingly wasteful) logic.

analyzer_plugin/lib/src/converter.dart
```diff
@@ -227,147 +239,68 @@ class HtmlTreeConverter {
...
 
-  List<AttributeInfo> _convertAttributes(ElementAst element) {
-    List<AttributeInfo> attributes = <AttributeInfo>[];
+  List<AttributeInfo> _convertAttributes({
+    List<ParsedAttributeAst> attributes: const [],
+    List<ParsedBananaAst> bananas: const [],
+    List<ParsedEventAst> events: const [],
+    List<ParsedPropertyAst> properties: const [],
+    List<ParsedReferenceAst> references: const [],
+    List<ParsedStarAst> stars: const [],
```
MichaelRFairhurst: I like the use of named params, makes it really clean when called

(there should probably be a lint for having more than x unnamed parameters, don't you think?)

analyzer_plugin/lib/src/converter.dart
```diff
@@ -227,147 +239,68 @@ class HtmlTreeConverter {
...
+          valueOffset,
+          dartParser.findMustaches(value, valueOffset),
+        ));
       }
     }
-    ;
 
-    element.events.forEach((event) {
-      attributes.add(_convertStatementsBoundAttribute(event, "(", ")"));
+    bananas.forEach((banana) {
```
MichaelRFairhurst: this could -- depending on preference -- be

```
bananas.map(_convertExpressionBoundAttribute).forEach(returnAttributes.add)
```

analyzer_plugin/lib/src/converter.dart
```diff
@@ -435,54 +368,28 @@ class HtmlTreeConverter {
...
-      origName =
-          (ast.prefixToken.errorSynthetic ? '' : ast.prefixToken.lexeme) +
-              ast.name +
-              (ast.postfix != null ? '.${ast.postfix}' : '') +
-              (ast.suffixToken.errorSynthetic ? '' : ast.suffixToken.lexeme);
-      origNameOffset = ast.prefixToken.offset;
-
-      value = ast.value;
+      ParsedEventAst ast) {
+    var origName =
```
MichaelRFairhurst: nested ternaries...hmm, should probably have at least one if statement I think

analyzer_plugin/test/abstract_angular.dart
MichaelRFairhurst: Looks like you did this, and github didn't realize. Commenting here so we know

deps/pubspec.yaml
```diff
@@ -6,7 +6,7 @@ dependencies:
   angular_ast:
     git:
       url: https://github.com/mk13/angular_ast
-      ref: master
+      ref: on_bind_parsing
```
MichaelRFairhurst: This can be master again, right?

analyzer_plugin/lib/src/converter.dart
mk13: Agreed, especially when the values have defaults - makes things so much cleaner.

deps/pubspec.yaml
mk13: Oh yeah, that's true. But I might have to switch this again.

analyzer_plugin/lib/src/converter.dart
mk13: I like it, changing it 👍

mk13:
I wanted to condense the three element type condensing - but that depends on the condensing of the three types in the angular_ast. Seeing as how it needs to be discussed more, it might be a good point to end this PR here and make the condensing of Element types into a separate PR. 

Ending the refactor of code/code clean-up and moving onto resolving the potential issue of errors not properly deserializing.

mk13:
Ping @MichaelRFairhurst .

Modified logic that loads the error codes into memory. Think we should add this logic here? or create a separate map within angular_ast?

MichaelRFairhurst:
Definitely worth building and running against greentea as a last check, but it LGTM!

PR #278, opened 2017-04-13
Resolve plain attributes -- in ranges, and strchecks for inputs.

MichaelRFairhurst:
Record resolved ranges for `x="y"` where `x` is an input or a constructor
`@Attribute`. If its an input, check that strings are assignable to it.
If not, give it its own error that hopefully explains the semi magical
situation.

### DIFF
--- a/analyzer_plugin/lib/ast.dart
+++ b/analyzer_plugin/lib/ast.dart
@@ -232,15 +232,17 @@ class DirectiveBinding {
 }
 
 /**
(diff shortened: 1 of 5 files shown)


MichaelRFairhurst:
@matanlurey @TedSander I'll leave this for a day for your review if you're interested. Just ask for more time if you need it, and let me know if you'll skip this one so I can submit.

Issue #281, opened 2017-04-19
The css unit % is not a valid css identifier


zoechi:
for
    
```html
[style.width.%]="100"
```
(at the % sign)

mk13:
Thanks! Will look into it. Will also add to the angular_ast.

PR #282, opened 2017-04-19
Percentage use fix in style.width and style.height. Test cases added

mk13:
`[style.width.%]` and `[style.height.%]` are allowed; but usage of % is a non width or height should still raise error.

Ref issue #281

### DIFF
--- a/analyzer_plugin/lib/src/resolver.dart
+++ b/analyzer_plugin/lib/src/resolver.dart
@@ -1211,7 +1211,11 @@ class SingleScopeResolver extends AngularScopeVisitor {
     if (dotpos != -1) {
       cssPropertyName = attribute.name.substring(0, dotpos);
       var cssUnitName = attribute.name.substring(dotpos + '.'.length);
-      if (!_isCssIdentifier(cssUnitName)) {
+      var validUnitName =
+          (cssPropertyName == 'width' || cssPropertyName == 'height') &&
+              cssUnitName == '%';
+      validUnitName = validUnitName || _isCssIdentifier(cssUnitName);
+      if (!validUnitName) {
         errorListener.onError(new AnalysisError(
             templateSource,
             attribute.nameOffset + dotpos + 1,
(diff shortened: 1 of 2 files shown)


mk13:
Need to add similar logic into parallel branch used for angular_ast merging.

analyzer_plugin/lib/src/resolver.dart
```diff
@@ -1211,7 +1211,11 @@ class SingleScopeResolver extends AngularScopeVisitor {
     if (dotpos != -1) {
       cssPropertyName = attribute.name.substring(0, dotpos);
       var cssUnitName = attribute.name.substring(dotpos + '.'.length);
-      if (!_isCssIdentifier(cssUnitName)) {
+      var validUnitName =
+          (cssPropertyName == 'width' || cssPropertyName == 'height') &&
```
MichaelRFairhurst: Hmm, this is too restrictive. Could be margin, marginTop, etc, I think border too...

mk13:
Should be more comprehensive now.

analyzer_plugin/lib/src/resolver.dart
```diff
@@ -1212,8 +1245,7 @@ class SingleScopeResolver extends AngularScopeVisitor {
       cssPropertyName = attribute.name.substring(0, dotpos);
       var cssUnitName = attribute.name.substring(dotpos + '.'.length);
       var validUnitName =
-          (cssPropertyName == 'width' || cssPropertyName == 'height') &&
-              cssUnitName == '%';
+          styleWithPercent.contains(cssPropertyName) && cssUnitName == '%';
```
zoechi: Is there a reason `%` is special-cased? What about all the other CSS units like https://www.w3.org/TR/css3-values/#lengths

mk13: I believe it follows javascript syntax for binding styles; ex: `style.<someProp>.%=20` where the property must have a percentage binding. Is `style.length.%=50` legal in javascript? If so I'll have to fix it.

MichaelRFairhurst:
Ideally we (eventually) don't special case any units, but are able to validate all css properties and units.

I'm wary though of trying to validate this at the moment, because of things like browser extensions. For instance, what if `-moz-opacity` supports `%`?

But I think at a certain point we have to take a leap. I am happy to start here, and ideally we can soon add validation for more units like `px` `em` etc.

zoechi:
I see. I just wanted to be sure to not get a warning when I use one of the other units.

PR #284, opened 2017-04-19
Remove contextRoot, won't be required in master soon, and breaks dev

MichaelRFairhurst:
dependent upon https://codereview.chromium.org/2828103002/


MichaelRFairhurst:
this is now ready

PR #285, opened 2017-04-20
Update test_reflective_loader, fix analyzer warnings&errors

MichaelRFairhurst:
Cannot use the version we were using before. And I'm not sure how these analysis errors made it through travis, but they are bad too

### DIFF
--- a/analyzer_plugin/test/angular_driver_test.dart
+++ b/analyzer_plugin/test/angular_driver_test.dart
@@ -22,14 +22,15 @@ import 'package:unittest/unittest.dart';
 import 'abstract_angular.dart';
 
 main() {
-  groupSep = ' | ';
-  defineReflectiveTests(AngularParseHtmlTest);
-  defineReflectiveTests(BuildStandardHtmlComponentsTest);
-  defineReflectiveTests(BuildUnitDirectivesTest);
-  defineReflectiveTests(BuildUnitViewsTest);
-  defineReflectiveTests(ResolveDartTemplatesTest);
-  defineReflectiveTests(ResolveHtmlTemplatesTest);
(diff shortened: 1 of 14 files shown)


analyzer_plugin/test/angular_driver_test.dart
```diff
@@ -2487,7 +2488,7 @@ class ComponentA {
 }
 ''';
     Source dartSource = newSource('/weird.dart', code);
-    Soruce htmlSource =
+    Source htmlSource =
```
liuming0: User var instead of Source.

server_plugin/test/analysis_test.dart
```diff
@@ -25,10 +25,11 @@ import 'package:unittest/unittest.dart';
 import 'mock_sdk.dart';
 
 main() {
-  groupSep = ' | ';
-  //defineReflectiveTests(AngularNavigationContributorTest);
-  //defineReflectiveTests(AngularOccurrencesContributorTest);
-  defineReflectiveTests(EmptyTest);
+  defineReflectiveSuite(() {
+    //defineReflectiveTests(AngularNavigationContributorTest);
+    //defineReflectiveTests(AngularOccurrencesContributorTest);
```
liuming0: Remove the //..

analyzer_plugin/test/angular_driver_test.dart
MichaelRFairhurst: good call

server_plugin/test/analysis_test.dart
MichaelRFairhurst: these actually have to be disabled :(

unless, is there a better way to skip them with whatever the heck test version we're using?

PR #288, opened 2017-04-21
Summary integration sync

mk13:
Resync to master and one bug fix.


mk13:
Tests will fail until change in angular_ast is pulled into master.

mk13:
This is correct; I'll probably have to do another one shortly.

PR #291, opened 2017-04-25
@ContentChild and @ContentChildren validation.

MichaelRFairhurst:
Validates:
- ElementRef
- TemplateRef
- Components
- Directives
- String-based let bindings (and checks property match)
- all others as errors
- property has QueryList assignability for ContentChildren
- ContentChild is only matched once

Changes:
- elements matching ContentChild of parent are counted as transcluded,
error wise
- DirectiveResolver is split into two phases, so that transclusions can
be checked _after_ inner content is directive resolved (for the sake of
ContentChild)
- added "StandardAngular" where we get certain special ClassElements
before running any other analysis, so that we can check if some random
type x is a supertype of angular's ElementRef.
- added ContentChildBindings so we can track the results of this
process to affect autocompletion suggestions, navigation, refactoring

Important edge cases:
- proper generic type checking for Iterable<T> as QueryList<T>
- using isSupertypeOf instead of isAssignableTo, because downcasting is
_known_ to be wrong (otherwise in dart downcasting is assumed
deliberate)
- don't supress transclusion errors for element A within B just because
some parent of B understands A as a contentChild -- has nothing to do
with A not belonging to B.
- don't report duplicate ContentChild errors when the duplicate is
inside a prior match (in fact, don't even report assignability errors!)
- directive matches require using exportAs with the #x="y" syntax.
- getting the "constant value" fields with inheritance requires using
`.getField("(super)")` the correct amount of times. This looks VERY
fragile to me.
- Native elements get Components, but they become ElementRefs not
TableElements. Ensure people don't use @ContentChild(TableElement).

Tests:
- changed to use 'package:angular2' instead of '/angular2/' because
that is required to build StandardAngular. However, not all tests are
changed, because its still ok to use '/angular2' too.
- added ElementRef and TemplateRef and QueryList classes.

Todos:
- ViewChild/ViewChildren
- track matched #attrs instead of just Elements
- revisit the highlighted ranges, can we make them more specific?
- instantiate classes to bounds

### DIFF
--- a/analyzer_plugin/lib/src/summary/format.dart
+++ b/analyzer_plugin/lib/src/summary/format.dart
@@ -750,6 +750,8 @@ class SummarizedDirectiveBuilder extends Object
   List<SummarizedBindableBuilder> _inputs;
   List<SummarizedBindableBuilder> _outputs;
   List<SummarizedDirectiveUseBuilder> _subdirectives;
+  List<SummarizedContentChildFieldBuilder> _contentChildFields;
+  List<SummarizedContentChildFieldBuilder> _contentChildrenFields;
 
   @override
   bool get isComponent => _isComponent ??= false;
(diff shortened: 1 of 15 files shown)


analyzer_plugin/lib/src/summary/format.dart
```diff
@@ -2006,3 +2093,185 @@ abstract class _SummarizedNgContentMixin implements idl.SummarizedNgContent {
...
+  /**
+   * Accumulate non-[informative] data into [signature].
+   */
+  void collectApiSignature(api_sig.ApiSignature signature) {
+    signature.addString(this._fieldName ?? '');
+    signature.addInt(this._nameOffset ?? 0);
+    signature.addInt(this._nameLength ?? 0);
+    signature.addInt(this._typeOffset ?? 0);
+    signature.addInt(this._typeLength ?? 0);
+  }
```
mk13: Could probably just use the getter methods since they're already doing null check and assignment of default values.

MichaelRFairhurst: This is actually generated code from the analyzer, produced based on `idl.dart`

PR #296, opened 2017-04-30
update dependencies to point to the right place

### LINKED ISSUE #295: Invalid dependency_overrides in angular_analyzer_plugin

daniel-v:
fixes #295 

There are some SDK packages that suffer from the same issue after running `tools/get_deps.sh`. This is only half of the fix.


MichaelRFairhurst:
Thanks Daniel, very glad to get your first PR! Did you want to do the other half of the fix? I might take it on otherwise, always good to keep the build green!

daniel-v:
Thank you! I'll create a PR to SDK in coming hour or so.

PR #297, opened 2017-05-01
Autocompletion fix

mk13:
Added support for autocompletion on .html files. Inline HTML found within .dart files still does not have support yet; will add them once plugin support is added. 

Breakdown: 
* Re-added directives into view.directives.
* Minor '.toList()' conversions to help with debugging.
* 'sendAngularCompletions()' is the primary function that will be called within the analysis_server to enable .html autocompletion.
* Contributor condensed into a single contributor; no longer separate ones for .dart-inline and .html.
* New base test case added for server tests: 'AbstractAngularTest' instead of 'AbstractAngularTaskTest'.

mock_sdk.dart is a copy and paste from analyzer component; no need for review.

### DIFF
--- a/analyzer_plugin/lib/src/angular_driver.dart
+++ b/analyzer_plugin/lib/src/angular_driver.dart
@@ -328,10 +328,11 @@ class AngularDriver
     return result;
   }
 
-  Future<Template> getTemplateForFile(String filePath) async {
+  Future<List<Template>> getTemplateForFile(String filePath) async {
+    var templates = <Template>[];
     var isDartFile = filePath.endsWith('.dart');
     if (!isDartFile && !filePath.endsWith('.html')) {
-      return null;
+      return templates;
     }
     var directiveResults = isDartFile
         ? await resolveDart(
@@ -351,11 +352,11 @@ class AngularDriver
             ? view.source.toString() == filePath
             : view.templateUriSource?.fullName == filePath;
         if (match) {
-          return view.template;
+          templates.add(view.template);
         }
       }
     }
-    return null;
+    return templates;
   }
 
   Future<DirectivesResult> resolveHtmlFrom(
--- a/analyzer_plugin/lib/src/angular_driver.dart
+++ b/analyzer_plugin/lib/src/angular_driver.dart
@@ -310,14 +310,16 @@ class AngularDriver
     }
 
     final summary = new LinkedHtmlSummaryBuilder()
-      ..errors = summarizeErrors(
-          result.errors.where((error) => error is! FromFilePrefixedError))
+      ..errors = summarizeErrors(result.errors
+          .where((error) => error is! FromFilePrefixedError)
+          .toList())
       ..errorsFromPath = result.errors
           .where((error) => error is FromFilePrefixedError)
           .map((error) => new SummarizedAnalysisErrorFromPathBuilder()
             ..path = (error as FromFilePrefixedError).fromSourcePath
             ..originalError =
-                summarizeError((error as FromFilePrefixedError).originalError));
+                summarizeError((error as FromFilePrefixedError).originalError))
+          .toList();
     final List<int> newBytes = summary.toBuffer();
     byteStore.put(key, newBytes);
 
@@ -399,7 +401,7 @@ class AngularDriver
           setIgnoredErrors(template, document);
           final resolver = new TemplateResolver(
               context.typeProvider,
-              standardHtml.components.values,
+              standardHtml.components.values.toList(),
               standardHtml.events,
               standardHtml.attributes,
               tplErrorListener);
--- a/analyzer_plugin/lib/src/angular_driver.dart
+++ b/analyzer_plugin/lib/src/angular_driver.dart
@@ -285,7 +285,10 @@ class AngularDriver
             source.exists() ? source.contents.data : "")(getSource(path));
   }
 
-  Future<DirectivesResult> resolveHtml(String htmlPath, {bool ignoreCache: false,}) async {
+  Future<DirectivesResult> resolveHtml(
+    String htmlPath, {
+    bool ignoreCache: false,
+  }) async {
     final key = getHtmlKey(htmlPath);
     final htmlSource = _sourceFactory.forUri("file:" + htmlPath);
     final List<int> bytes = byteStore.get(key);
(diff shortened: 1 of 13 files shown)


server_plugin/lib/src/completion.dart
```diff
@@ -197,37 +198,33 @@ class ReplacementRangeCalculator extends AngularAstVisitor {
...
+   * for the given completion [request].
    */
   Future<List<CompletionSuggestion>> computeSuggestions(
       CompletionRequest request) async {
-    if (request.source.shortName.endsWith('.html')) {
-      //return new TemplateCompleter().computeSuggestions(
-      //    request, templates, standardHtmlEvents, standardHtmlAttributes);
-    }
+    var filePath = request.source.toString();
+    var events = driver.standardHtml.events.values;
```
MichaelRFairhurst: What do you think...my impression here is that this should probably `await driver.getStandardHtml()`, just in case a completion is requested before standandHtml is ready. Is that possible? should we add an assertion that `driver.standardHtml` is not null? Make it private?

analyzer_plugin/lib/src/angular_driver.dart
```diff
@@ -309,20 +312,52 @@ class AngularDriver
...
+      return null;
+    }
+    for (var directive in directives) {
+      if (directive is Component) {
+        var view = directive.view;
+        var match = isDartFile
+            ? view.source.toString() == filePath
+            : view.templateUriSource?.fullName == filePath;
+        if (match) {
+          return view.template;
```
MichaelRFairhurst: For html file cases, we want to support having two autocomplete contexts.

For instance, one common example is having both a production and a test component definition that reference the same dart file. If we only suggested the production completions, that would probably be OK, but if we only suggested the test completions, that would likely mean suggesting like, half of the member variables, maybe half of the directives, etc.

Of course, distinguishing the two is pretty hard. So it should probably be a list of anything that matches this condition.

server_plugin/lib/src/completion.dart
```diff
@@ -236,118 +233,115 @@ class TemplateCompleter {
 
   Future<List<CompletionSuggestion>> computeSuggestions(
       CompletionRequest request,
-      List<Template> templates,
+      Template template,
```
MichaelRFairhurst: see my comments on `getTemplateForFile`, probably still want to use a List of templates.

Though I do really like that there's less indentation in this function! maybe a function that calls this one?

server_plugin/lib/src/completion.dart
```diff
@@ -236,118 +233,115 @@ class TemplateCompleter {
...
+      suggestOutputs(target.parent.boundDirectives, suggestions,
+          standardHtmlEvents, target.parent.boundStandardOutputs);
+    } else if (target is TextAttribute) {
+      suggestInputs(target.parent.boundDirectives, suggestions,
+          standardHtmlAttributes, target.parent.boundStandardInputs);
+      suggestOutputs(target.parent.boundDirectives, suggestions,
+          standardHtmlEvents, target.parent.boundStandardOutputs);
+    } else if (target is TextInfo) {
+      suggestHtmlTags(template, suggestions);
+      suggestTransclusions(target.parent, suggestions);
```
MichaelRFairhurst: Hard to review this because of indentation changes. Should I read over it for changes or is it mostly the same?

mk13: Interesting, github decided to see these all as deltas. The main 'if-else' chain block shouldn't have changed.

analyzer_plugin/lib/src/resolver.dart
```diff
@@ -119,6 +119,18 @@ class TemplateResolver {
     var allDirectives = <AbstractDirective>[]
       ..addAll(standardHtmlComponents)
       ..addAll(view.directives);
+
+    for (var directive in view.directives) {
+      if (directive.elementTags != null && directive.elementTags.isNotEmpty) {
```
mk13: Sure, sounds like a plan.

analyzer_plugin/lib/src/view_extraction.dart
```diff
@@ -173,14 +173,6 @@ class ViewExtractor extends AnnotationProcessorMixin {
                   element.variable.constantValue != null) {
             directiveReferences.add(new DirectiveReference(
                 name, prefix, new SourceRange(item.offset, item.length)));
-            //DartObject value = element.variable.constantValue;
```
mk13: Pre-summary changes, it looks like this code was leftover that was primarily responsible for loading the directives and elementTags(view level). The directives are being added in 'directive_linking.dart' and elementTags are being add to the view level with lazy calculation. 

Doesn't seem like there's a purpose for this commented out section of code.

analyzer_plugin/lib/starter.dart
```diff
@@ -15,6 +25,8 @@ class Starter {
     ContextBuilder.onCreateAnalysisDriver = onCreateAnalysisDriver;
     server.onResultErrorSupplementor = sumErrors;
     server.onNoAnalysisResult = sendHtmlResult;
+    server.onNoAnalysisCompletion = sendAngularCompletions;
+    //server.onExtraCompletionContributor = sendAngularContributor;
```
mk13: Actually I'm just going to remove it - it shouldn't be used once Plugin support is implemented.

analyzer_plugin/lib/starter.dart
```diff
@@ -69,4 +81,69 @@ class Starter {
...
+              null,
+              null,
+              server.resourceProvider,
+              server.searchEngine,
+              source,
+              params.offset,
+              performance,
+              server.ideOptions);
+          completionHandler.setNewRequest(completionRequest);
+          server.sendResponse(new CompletionGetSuggestionsResult(completionId)
```
mk13: NotificationManager doesn't have full support for completion yet: 
```
    errors = new ResultCollector<List<server.AnalysisError>>(serverId,
        predicate: _isIncluded);
    folding = new ResultCollector<List<server.FoldingRegion>>(serverId);
    highlights = new ResultCollector<List<server.HighlightRegion>>(serverId);
    navigation = new ResultCollector<server.AnalysisNavigationParams>(serverId);
    occurrences = new ResultCollector<List<server.Occurrences>>(serverId);
    outlines = new ResultCollector<List<server.Outline>>(serverId);
```
These are the only response types supported; looks like server.sendResponse is a direct way to communicate against the wire/channel.

analyzer_plugin/lib/starter.dart
```diff
@@ -69,4 +81,69 @@ class Starter {
...
+
+    if (server.contextManager.isInAnalysisRoot(filePath)) {
+      for (final driverPath in angularDrivers.keys) {
+        if (server.contextManager.getContextFolderFor(filePath).path ==
+            driverPath) {
+          final driver = angularDrivers[driverPath];
+
+          var completionContributor = new AngularCompletionContributor(driver);
+          CompletionRequestImpl completionRequest = new CompletionRequestImpl(
+              null,
```
mk13: Done

analyzer_plugin/lib/src/angular_driver.dart
mk13: Wouldn't the source provided by the Completion Request object suffice to distinguish which of the test/production component to complete for?

MichaelRFairhurst: No, it can be one template shared by two dart files.

```
foo.dart
foo.html
foo_test.dart
```

where `foo_test.dart` implements half the members of `foo.dart`, loads half the directives, etc.

analyzer_plugin/lib/src/view_extraction.dart
MichaelRFairhurst: Oh, I see, it was already commented out, and you deleted it.

OK good! Won't have confusion again like what I just had, haha!

analyzer_plugin/lib/starter.dart
MichaelRFairhurst: dang! OK

server_plugin/lib/src/completion.dart
MichaelRFairhurst: Cool. And it definitely is nice to have less indentation!

analyzer_plugin/lib/src/angular_driver.dart
mk13: Ohhh I see what you're saying. Got it.

MichaelRFairhurst:
Still LGTM

PR #299, opened 2017-05-02
Implement new methods that will soon be part of AnalysisDriverGeneric

MichaelRFairhurst:


analyzer_plugin/lib/src/angular_driver.dart
```diff
@@ -72,6 +72,23 @@ class AngularDriver
...
+   */
+  void setPriorityFiles(List<String> priorityPaths) {
+    // TODO analyze these files first
+  }
+
+  /**
+    * Notify the driver that the client is going to stop using it.
+    */
+  void dispose() {
+    // TODO anything we need to do here?
```
bwilkerson: As per external discussion, this eventually needs to remove the driver from the scheduler.

MichaelRFairhurst: Will do. Can you ping me when that becomes public (and or another solution comes up)?

PR #301, opened 2017-05-02
Prevent inf loop of trying to resolve standardAngular in workspaces t…

mk13:
Tests will fail until PR #300 is pushed.


MichaelRFairhurst:
I will solve those test failures in a bit

PR #303, opened 2017-05-04
Plain attribute autocompletion

mk13:
First iteration of plain text attribute completion.

Doesn't include global angular attributes yet.

### DIFF
--- a/playground/lib/overview_component.html
+++ b/playground/lib/overview_component.html
@@ -9,4 +9,4 @@
 
 <my-counter [count]='4' (incremented)='items.add($event.toString())'></my-counter>
 
-<my-counter></my-counter>
+<my-counter ></my-counter>
--- a/playground/lib/overview_component.html
+++ b/playground/lib/overview_component.html
@@ -9,4 +9,4 @@
 
 <my-counter [count]='4' (incremented)='items.add($event.toString())'></my-counter>
 
-<my-counter ></my-counter>
+<my-counter></my-counter>
--- a/server_plugin/lib/src/completion.dart
+++ b/server_plugin/lib/src/completion.dart
@@ -424,6 +424,13 @@ class TemplateCompleter {
         }
         suggestions.add(_createInputSuggestion(input, DART_RELEVANCE_DEFAULT,
             _createInputElement(input, protocol.ElementKind.SETTER)));
+        if (input.setterType is String) {
+          suggestions.add(_createPlainAttributeSuggestions(
+              input,
+              DART_RELEVANCE_DEFAULT,
+              _createPlainAttributeElement(
+                  input, protocol.ElementKind.SETTER)));
+        }
       }
     }
 
@@ -438,6 +445,12 @@ class TemplateCompleter {
       }
       suggestions.add(_createInputSuggestion(input, DART_RELEVANCE_DEFAULT - 1,
           _createInputElement(input, protocol.ElementKind.SETTER)));
+      if (input.setterType is String) {
+        suggestions.add(_createPlainAttributeSuggestions(
+            input,
+            DART_RELEVANCE_DEFAULT,
+            _createPlainAttributeElement(input, protocol.ElementKind.SETTER)));
+      }
     }
   }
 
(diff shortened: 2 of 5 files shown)


playground/lib/overview_component.html
```diff
@@ -9,4 +9,4 @@ <h1 #h1>Showing {{items.length}} items:</h1>
 
 <my-counter [count]='4' (incremented)='items.add($event.toString())'></my-counter>
 
-<my-counter></my-counter>
\ No newline at end of file
+<my-counter ></my-counter>
```
MichaelRFairhurst: Probably best to remove this change from history. I think you can do

```
git reset
git checkout playground/overview_component.html
git commit --amend
```

and then push `--force` or it will complain.

Not sure the exact steps.

server_plugin/lib/src/completion.dart
```diff
@@ -424,6 +424,13 @@ class TemplateCompleter {
         }
         suggestions.add(_createInputSuggestion(input, DART_RELEVANCE_DEFAULT,
             _createInputElement(input, protocol.ElementKind.SETTER)));
+        if (input.setterType.displayName == 'String') {
```
MichaelRFairhurst: This will exclude "dynamic" such as material component's `raised`. If you can get the `context`, you can get its `typeProvider` and check if `typeProvider.stringType.isAssignableTo(input.setterType)`, which will be true for inputs of type `dynamic`, `Object`, `String`, etc.

server_plugin/lib/src/completion.dart
```diff
@@ -424,6 +424,13 @@ class TemplateCompleter {
...
+          suggestions.add(_createPlainAttributeSuggestions(
+              input,
+              DART_RELEVANCE_DEFAULT,
+              _createPlainAttributeElement(
+                  input, protocol.ElementKind.SETTER)));
+        }
```
MichaelRFairhurst: by adding this block to `suggestInputs`, it will show up when people are editing `[foo|`...what are your thoughts on that? We don't for instance currently suggest outputs in this case...And I've wondered if that was the right choice.

What do you think? Is the experience wrong when you test this by editing/autocompleting a bracket-based binding?

server_plugin/lib/src/completion.dart
```diff
@@ -438,6 +445,12 @@ class TemplateCompleter {
       }
       suggestions.add(_createInputSuggestion(input, DART_RELEVANCE_DEFAULT - 1,
           _createInputElement(input, protocol.ElementKind.SETTER)));
+      if (input.setterType.displayName == 'String') {
+        suggestions.add(_createPlainAttributeSuggestions(
+            input,
+            DART_RELEVANCE_DEFAULT - 1,
```
MichaelRFairhurst: good catch on relevance!

This makes me wonder, should we suggest string bindings at lower relevance overall do you think?

And regarding my other comment about `dynamic` etc, maybe we could even recommend string bindings at a lower priority unless the type is exactly `String`?

server_plugin/test/completion_contributor_test.dart
```diff
@@ -1289,6 +1345,57 @@ class OtherComp {
...
+
+    await resolveSingleTemplate(dartSource);
+    await computeSuggestions();
+    expect(replacementOffset, completionOffset);
+    expect(replacementLength, 0);
+    assertSuggestSetter('name');
+    assertNotSuggested('intInput');
+    assertSuggestSetter('id', relevance: DART_RELEVANCE_DEFAULT - 1);
+  }
+
```
MichaelRFairhurst: might be worth adding a test where the plain attribute is started ie `<my-tag foo^></mytag>`

server_plugin/lib/src/completion.dart
mk13: Hmm this is a good point. It appears seamless on my IDE, but that's because IntelliJ is removing the plain-text suggestions by itself; this seems hackish though. I can move this to a separate logic block.

server_plugin/lib/src/completion.dart
mk13: Sure that sounds good. However, I don't think that string-bindings should be lower priority just by itself; it should hold similar priority to its bracket-counterpart. For the dynamic part, I think it would be good to put those at lower priority. 

I can set the priority as following: 
1) Input-binding with bracket == Input-binding without bracket AND is string
2) Input-binding without bracket that is non-String
3) StandardHtml-input-binding with bracket == .. without bracket AND is string
4) StandardHtml-input-binding without bracket that is non-String

server_plugin/test/completion_contributor_test.dart
mk13: Good point, will add.

server_plugin/lib/src/completion.dart
MichaelRFairhurst: seems like a great order to me.

mk13:
Ping @MichaelRFairhurst ; significant changes made - potentially need re-review.

server_plugin/lib/src/completion.dart
```diff
@@ -244,10 +255,12 @@ class TemplateCompleter {
...
   Future<List<CompletionSuggestion>> computeSuggestions(
-      CompletionRequest request,
-      Template template,
-      List<OutputElement> standardHtmlEvents,
-      List<InputElement> standardHtmlAttributes) async {
+    CompletionRequest request,
+    Template template,
+    List<OutputElement> standardHtmlEvents,
+    List<InputElement> standardHtmlAttributes,
+    TypeProvider typeProvider,
```
MichaelRFairhurst: I think you can get the `TypeProvider` off of the template like so: `template.view.component.classElement.enclosingElement.enclosingElement.context.typeProvider`

mk13: Oh wow... but cool. At least we can remove some overhead by not having to do another function call.

PR #310, opened 2017-05-09
Autocomplete stars

MichaelRFairhurst:
Complete inputs inside of template elements  …
Complete, for instance, `trackBy` inside of an ngFor.

Analyze and suggest directives that appear to want `*star` syntax.  …
If a `TemplateRef` is in the constructor, it works with star syntax.
Warn when not used with star syntax, and warn when star syntax is used
on a directive that does not have this special ctor as otherwise the
template will go unused.

Its not a foolproof system, I don't think. So detect that separately
from detection of ngIf and ngFor without stars. Especially so that in
the cases where this goes wrong, users can opt out of the error without
losing errors for those core directives' validation...and give
different messages.

Not sure what's more likely...that this expects too many things to be
used with stars, or not enough...

Also suggest this in autocomplete for great profit. Assume that these
directives rely on a property binding, so suggest *p for all p which
are property selectors.

Note this is also a less than perfect system, as it includes suggesting
`*ngForOf`, since that's part of `NgFor`s selector. Hardcode that out,
at least for now.

With this plus suggesting `trackBy:`, we now (I think) suggest
everything relevant for two way autocomplete except `template=`
attributes themselves and the `let` keyword. Wanted to do the latter
but it seemed like it might be a bit treacherous: think `ngIf="^"` vs
`ngFor="^"`, it would be the first time we'd suggest both dart and
non-dart things at once.

### DIFF
--- a/analyzer_plugin/lib/ast.dart
+++ b/analyzer_plugin/lib/ast.dart
@@ -231,10 +231,10 @@ abstract class HasDirectives {
 /// Naming here is important: "bound directive" != "directive binding."
 class DirectiveBinding {
   final AbstractDirective boundDirective;
-  final List<InputBinding> inputBindings = [];
-  final List<OutputBinding> outputBindings = [];
-  final Map<ContentChild, ContentChildBinding> contentChildBindings = {};
-  final Map<ContentChild, ContentChildBinding> contentChildrenBindings = {};
+  final inputBindings = <InputBinding>[];
+  final outputBindings = <OutputBinding>[];
+  final contentChildBindings = <ContentChild, ContentChildBinding>{};
+  final contentChildrenBindings = <ContentChild, ContentChildBinding>{};
 
   DirectiveBinding(this.boundDirective);
 }
--- a/analyzer_plugin/lib/ast.dart
+++ b/analyzer_plugin/lib/ast.dart
@@ -97,11 +97,15 @@ class TemplateAttribute extends BoundAttributeInfo implements HasDirectives {
   final boundStandardOutputs = <OutputBinding>[];
   @override
   final boundStandardInputs = <InputBinding>[];
+
   List<AbstractDirective> get directives =>
       boundDirectives.map((bd) => bd.boundDirective).toList();
 
+  String prefix;
+
   TemplateAttribute(String name, int nameOffset, String value, int valueOffset,
-      String originalName, int originalNameOffset, this.virtualAttributes)
+      String originalName, int originalNameOffset, this.virtualAttributes,
+      {this.prefix})
       : super(name, nameOffset, value, valueOffset, originalName,
             originalNameOffset);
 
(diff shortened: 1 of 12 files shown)


analyzer_plugin/lib/ast.dart
```diff
@@ -91,14 +91,17 @@ abstract class BoundAttributeInfo extends AttributeInfo {
 
 class TemplateAttribute extends BoundAttributeInfo implements HasDirectives {
   final List<AttributeInfo> virtualAttributes;
-  List<DirectiveBinding> boundDirectives = <DirectiveBinding>[];
-  List<OutputBinding> boundStandardOutputs = <OutputBinding>[];
-  List<InputBinding> boundStandardInputs = <InputBinding>[];
+  final List<DirectiveBinding> boundDirectives = <DirectiveBinding>[];
```
liuming0: Remove "List<DirectiveBinding>" since you are using final. Same for the rest.

analyzer_plugin/lib/src/resolver.dart
```diff
@@ -751,10 +754,22 @@ class DirectiveResolver extends AngularAstVisitor {
             directive.contentChildren.isNotEmpty) {
           outerBindings.add(binding);
         }
+
+        // Specifically exclude NgIfand NgFor, they have their own error since
```
liuming0: Missing a space, i.e. NgIf and NgFor

server_plugin/lib/src/completion.dart
```diff
@@ -495,6 +525,33 @@ class TemplateCompleter {
     }
   }
 
+  suggestInputsInTemplate(
+      TemplateAttribute templateAttr, List<CompletionSuggestion> suggestions,
+      {AttributeInfo currentAttr}) {
+    List<DirectiveBinding> directives = templateAttr.boundDirectives;
```
liuming0: var

server_plugin/lib/src/completion.dart
```diff
@@ -495,6 +525,33 @@ class TemplateCompleter {
...
+    for (DirectiveBinding binding in directives) {
+      Set<InputElement> usedInputs = new HashSet.from(binding.inputBindings
```
liuming0: var

server_plugin/lib/src/completion.dart
```diff
@@ -495,6 +525,33 @@ class TemplateCompleter {
...
+          .where((b) => b.attribute != currentAttr)
+          .map((b) => b.boundInput)).toSet();
+
+      for (InputElement input in binding.boundDirective.inputs) {
+        // don't recommend [name] [name] [name]
```
liuming0: var

server_plugin/lib/src/completion.dart
```diff
@@ -662,4 +764,25 @@ class TemplateCompleter {
     return new protocol.Element(kind, name, flags,
         location: location, returnType: outputElement.eventType.toString());
   }
+
+  CompletionSuggestion _createStarAttrSuggestion(AttributeSelector selector,
+      int defaultRelevance, protocol.Element element) {
+    String completion = '*' + selector.nameElement.name;
```
liuming0: var

server_plugin/lib/src/completion.dart
```diff
@@ -662,4 +764,25 @@ class TemplateCompleter {
...
+    return new CompletionSuggestion(CompletionSuggestionKind.IDENTIFIER,
+        defaultRelevance, completion, completion.length, 0, false, false,
+        element: element);
+  }
+
+  protocol.Element _createStarAttrElement(
+      AttributeSelector selector, protocol.ElementKind kind) {
+    String name = '*' + selector.nameElement.name;
```
liuming0: var, and all the rest in this method.

mk13:
If the lints are fairly substantial might be worth it to clean those up with this PR? Or separate PR?

PR #311, opened 2017-05-09
Fix for source; null check added

mk13:
Fix for: https://github.com/dart-lang/sdk/issues/29576#issuecomment-299958547

I've been trying to repro the issue by removing source imports before parsing - but haven't been able to do so. It's odd that the 'type'(ClassElement) could be found but the 'source' is showing up as null. 

This is a temporary fix in place, but it might cause strange errors to appear in the future since the directive-linking is affected. With this solution at least it won't cause a full-on crash.

### DIFF
--- a/analyzer_plugin/lib/src/directive_linking.dart
+++ b/analyzer_plugin/lib/src/directive_linking.dart
@@ -226,7 +226,7 @@ class ChildDirectiveLinker implements DirectiveMatcher {
             new StringToken(TokenType.IDENTIFIER, reference.name, 0)),
         null);
 
-    if (type != null) {
+    if (type != null && type.source != null) {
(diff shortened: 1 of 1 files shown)


MichaelRFairhurst:
I think to repro you'll want to remove angular2 from your `.packages` file....I'm also not sure why the type wouldn't be null but would have a source...

mk13:
I'll try that out; and that's the strange part - the type is perfectly fine but the source is source is appearing as null.

MichaelRFairhurst:
we might need to get more details from gunter

mk13:
Removing angular2 from .packages didn't recreate the issue. I've also tried hard-removing the direct files while keeping the path in .packages, but that's not causing the issue either. I'll post in the original issue to get some more info from Gunter.

PR #314, opened 2017-05-10
Remove searchEngine from CompletionRequest reflecting changes made in…

mk13:
Reflecting changes made in completion core; removal of 'searchEngine'.


server_plugin/test/completion_contributor_test_util.dart
```diff
@@ -548,7 +546,6 @@ abstract class BaseCompletionContributorTest extends AbstractAngularTest {
   void setUp() {
     super.setUp();
     index = createMemoryIndex();
```
bwilkerson: My guess is that you don't need the `index` field any more.

PR #315, opened 2017-05-11
Cleaned up unused imports and unused override reflecting recent changes

mk13:


server_plugin/lib/src/embedded_dart_completion_request.dart
```diff
@@ -21,25 +19,20 @@ class EmbeddedDartCompletionRequest implements DartCompletionRequest {
     request.checkAborted();
 
     Source libSource;
-    if (request.context != null) {
```
MichaelRFairhurst: this should be changed to match https://codereview.chromium.org/2877653002/patch/20001/30005 as closely as possible. Looks like the check and assignment to `libSource` was removed, but they do find `coreLib` which we probably need to do as well

server_plugin/lib/src/embedded_dart_completion_request.dart
```diff
@@ -64,6 +54,7 @@ class EmbeddedDartCompletionRequest implements DartCompletionRequest {
       opType.includeVoidReturnSuggestions = false;
     }
 
+    // Below is copied from analysis_server.../completion_manager.dart.
```
MichaelRFairhurst: thanks for adding this comment....was fun to be confused over my own code, hopefully won't happen in this file again any time soon :)

PR #316, opened 2017-05-11
Get new plugin architecture in a workable development stage

MichaelRFairhurst:
See readme for how to get set up....currently doesn't do anything,
though.

Dependent upon https://codereview.chromium.org/2877863002


analyze_angular/lib/analyze_angular.dart
```diff
@@ -0,0 +1 @@
+// empty
```
mk13: Is the 'analyze_angular' going to be the new 'master' folder for the new architecture while we slowly move contents from analyzer_plugin and server_plugin into here?

MichaelRFairhurst: no, this is a temp package that "publishes" the analyzer plugin and does nothing else. To release, we'll move `tool/` into `angular2` itself, so this is mostly just a placeholder.

PR #324, opened 2017-05-15
Final sync of angular ast

mk13:
Sync up to master and merging into MASTER this time.


analyzer_plugin/test/resolver_test.dart
```diff
@@ -2669,29 +2690,29 @@ class TestPanel {
...
-</li>
-<li *ngFor="let item3 of items; let i=index" {{lowerEl}}>
-  {{item3.length}}
-</li>
-<div #lowerEl></div>
-""");
-    await _resolveSingleTemplate(dartSource);
-    errorListener.assertNoErrors();
-  }
+//  Future test_ngFor_variousKinds_useLowerIdentifier() async {
```
liuming0: Do you want to remove this part for good?

mk13: There were hints that this syntax would be deprecated. Codebase does not reveal any usage of this syntax, but want to get a firm confirmation that this format will no longer be supported before fully removing this test case incase it is not deprecated.

PR #325, opened 2017-05-16
Load angular drivers in new plugin, attempt errors+completion

MichaelRFairhurst:


analyzer_plugin/lib/src/angular_driver.dart
```diff
@@ -53,8 +51,8 @@ class AngularDriver
   final lastSignatures = <String, String>{};
   bool _hasAngularImported = false;
 
-  AngularDriver(this.server, this.dartDriver, this._scheduler, this.byteStore,
-      SourceFactory sourceFactory, this._contentOverlay) {
+  AngularDriver(this.notificationManager, this.dartDriver, this._scheduler,
```
bwilkerson: I really like your approach of creating an interface that we can implement on both sides so that we can avoid copying this code!

That said, I don't have a file named `notification_manager.dart` in the version of `angular_analyzer_plugin` that I pulled this afternoon, so the file might need to be added.

new_plugin/lib/plugin.dart
```diff
@@ -1,20 +1,22 @@
 // Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
+import 'dart:async';
 
+import 'package:analysis_server/src/services/completion/completion_core.dart';
+import 'package:analysis_server/src/services/completion/dart/completion_manager.dart';
```
bwilkerson: I should have added a TODO comment to remind us that we need to get an API in 'analyzer_plugin' that supports completions. These references to 'analysis_server' can't stay.

new_plugin/lib/plugin.dart
```diff
@@ -29,14 +31,26 @@ class AngularAnalysisPlugin extends ServerPlugin {
...
   AnalysisDriverGeneric createAnalysisDriver(plugin.ContextRoot contextRoot) {
-    ContextRoot root = new ContextRoot(contextRoot.root, contextRoot.exclude);
-    DartSdkManager sdkManager =
-        new DartSdkManager('/Users/brianwilkerson/Dev/dart/dart-sdk', true);
-    ContextBuilder builder =
-        new ContextBuilder(resourceProvider, sdkManager, null);
-    builder.analysisDriverScheduler = analysisDriverScheduler;
-    return builder.buildDriver(root);
+    final root = new ContextRoot(contextRoot.root, contextRoot.exclude);
+    final sdkManager = new DartSdkManager('/home/mfairhurst/.dart-sdk', true);
```
bwilkerson: We need to figure out the right way to find the SDK. Perhaps a TODO here as well.

new_plugin/lib/plugin.dart
```diff
@@ -29,14 +31,26 @@ class AngularAnalysisPlugin extends ServerPlugin {
...
+    final builder = new ContextBuilder(resourceProvider, sdkManager, null)
+      ..analysisDriverScheduler = analysisDriverScheduler;
+    final dartDriver = builder.buildDriver(root);
+
+    final sourceFactory = dartDriver.sourceFactory;
+
+    return new AngularDriver(
+        new ChannelNotificationManager(channel),
```
bwilkerson: I don't see a definition of `ChannelNotificationManager`. Did you forget to add a file, or am I missing something?

new_plugin/lib/plugin.dart
```diff
@@ -64,6 +78,23 @@ class AngularAnalysisPlugin extends ServerPlugin {
   }
 
   @override
+  void contentChanged(String path) {
+    for (final AngularDriver driver in driverMap.values) {
+      driver
+        ..addFile(path)
```
bwilkerson: I don't think we want to add the file. Given that this only gets invoked when the content of an existing file changes, adding it seems wrong. Also, we probably want to check (here or in the driver) to ensure that the file has been analyzed before we mark it as changed (but I don't remember off hand what `fileChanged` does).

new_plugin/lib/plugin.dart
MichaelRFairhurst: Yep! Oops! Added.

analyzer_plugin/lib/src/angular_driver.dart
MichaelRFairhurst: Oops, yes! Added.

MichaelRFairhurst: And yeah, funny enough, it hasn't needed an `AnalysisServer` for quite a while. Have been meaning to change it to `NotificationManager`, looks like I got my chance, but just with a slightly different class of the same name! :)

new_plugin/lib/plugin.dart
MichaelRFairhurst: I looked through `analyzer_plugin` for an API when files are added. I didn't see one, and it looked to me like this method was called in either case.

analyzer_plugin/lib/starter.dart
```diff
@@ -124,3 +125,18 @@ class Starter {
...
+class ServerNotificationManager implements NotificationManager {
+  final AnalysisServer server;
+
+  ServerNotificationManager(this.server);
+
+  @override
+  void recordAnalysisErrors(
+          String path, LineInfo lineInfo, List<AnalysisError> analysisErrors) =>
+      server.notificationManager.recordAnalysisErrors(
+          "angular driver",
```
mk13: Since we did a massive lint clean up recently, prefer single quote over double.

MichaelRFairhurst: ah good catch thanks!!

new_plugin/lib/plugin.dart
bwilkerson: It's possible that `AngularDriver` is different, but in `AnalysisDriver`, the method `addFile` is used to indicate that the file is to be "explicitly" analyzed (as opposed to "implicitly" analyzed). An explicitly analyzed file is one that always needs to be re-analyzed on changes; an implicitly analyzed file is one that is referenced by an analyzed file and only needs to be re-analyzed on changes if it continues to be referenced.

More intuitively, the files in the context root (the directory that the user opened) are explicitly analyzed (because the user explicitly opened them) and all other files are implicitly analyzed.

The intent of this distinction is to reduce the amount of work that needs to be done when something changes that might affect the analysis results.

The problem is that the content of implicitly analyzed files can also change. So, in `AnalysisDriver`, this would result in files that should be implicitly analyzed getting marked as files that should be explicitly analyzed, which could hurt performance. As I said, that might not be the case here; I was just assuming that you preserved that semantic.

MichaelRFairhurst: Hmm, we use "add file"

- to track which files might need to be reanalyzed by later changed files
- to reanalyze all added files

and change file

- to reanalyze all added files

Issue #329, opened 2017-05-17
`[readonly]` causes complaints about `readOnly`


zoechi:
![image](https://cloud.githubusercontent.com/assets/405837/26146012/f72e65ee-3aee-11e7-99a0-3e90f971a468.png)

```dart
class StringInputComponent extends DefaultValueAccessor implements OnChanges {
  NgModel _ngModel;
  String value;

  @Input()
  String label;

  @Input()
  bool readonly = false;
  ...
```

zoechi:
Almost allof my input elements have a `readonly` property and I get this warning now for each of them (since todays plugin update).

Dart VM version: 1.24.0-edge.6c8733dcbd0fe2b1edbec5f9138e266c9828e7f3 (Tue May 16 19:19:16 2017) on "linux_x64"

mk13:
Ah I know the reason for this, I'll path this right away since this is a relatively quick fix.

mk13:
Resolved now; PR #343

Issue #328, opened 2017-05-16
Dedupe suggestions based on directive.classElement


mk13:
Ex: NgIf appearing twice if NgIf and CORE_DIRECTIVES imported.

MichaelRFairhurst:
should be able to just dedupe on `d1 == d2`, `AbstractDirective` overloads `==` to check the underlying `classElement`.

Alternatively, we can put them into a `Set` instead of a `List` so they dedupe themselves.

mk13:
I used the set method :)

mk13:
Resolved for inputs PR #343, but will keep open since this will be persistent once attribute-directive will need deduping as well.

MichaelRFairhurst:
I think we should dedupe this right at the source:

```
directives: [NgIf, CORE_DIRECTIVES]
```

this should result in a `Component` with `.view.directives` of `[NgIf,
NgFor, NgForm]`, no duplicates. We should be able to do this by just making
`View.directives` a `Set`.

Otherwise, everything will be duplicated. For instance, `<div *ngIf="1">`
will probably show two errors, send two navigation items, report two usages
in the search, perhaps mess up refactoring......

mk13:
That sounds like a good plan.

mk13:
Done.

PR #343, opened 2017-05-17
Issue 329 - Custom defined for alternative names no longer trigger errors. Deduping added.

mk13:
Primarily alternative names for: 

- class : className
- innerHtml : innerHTML
- readonly : readOnly
- tabindex : tabIndex

Deduping also added because new changes will suggest inputs multiple times. 

Resolves issue #329 and issue #328


analyzer_plugin/lib/src/converter.dart
```diff
@@ -22,15 +22,6 @@ class HtmlTreeConverter {
...
 
-  // Following Angular2 Logic:
-  // https://github.com/dart-lang/angular2/blob/8220ba3a693aff51eed33cd1ec9542bde9017423/lib/src/compiler/schema/dom_element_schema_registry.dart#L199
-  static const attrToPropMap = const {
-    'class': 'className',
-    'innerHtml': 'innerHTML',
-    'readonly': 'readOnly',
-    'tabindex': 'tabIndex',
-  };
-
```
mk13: Resolving this at the converter level proved to be too shallow of a solution. New solution will now occur at the InputElement level.

analyzer_plugin/test/resolver_test.dart
```diff
@@ -378,6 +378,43 @@ class TestPanel {
...
+''');
+    final code = r'''
+<my-tag [readonly]="'blah'"></my-tag>
+''';
+    _addHtmlSource(code);
+    await _resolveSingleTemplate(dartSource);
+    errorListener.assertNoErrors();
+  }
+
+  // ignore: non_constant_identifier_names
```
liuming0: Use "///" for document comment

analyzer_plugin/test/resolver_test.dart
```diff
@@ -378,6 +378,43 @@ class TestPanel {
...
+    final code = r"""
+<span [class]='text'></span>
+""";
+    await angularDriver.getStandardHtml();
+    _addHtmlSource(code);
+    await _resolveSingleTemplate(dartSource);
+    errorListener.assertNoErrors();
+  }
+
+  // ignore: non_constant_identifier_names
```
liuming0: Use "///" for document comment

analyzer_plugin/test/resolver_test.dart
mk13: This isn't for document comment but to silence the lint error warning.

server_plugin/lib/src/completion.dart
```diff
@@ -702,12 +704,26 @@ class TemplateCompleter {
         location: location);
   }
 
-  CompletionSuggestion _createInputSuggestion(InputElement inputElement,
+  List<CompletionSuggestion> _createInputSuggestion(InputElement inputElement,
```
MichaelRFairhurst: should be pluralized.

Another option: do we want to choose whether to even suggest the nonalternative names at all? That is, if `[class]` works, and is preferred, should we even suggest `[className]`?

If we make `alternativeNames` into `alternativeName`, and only suggest that since that's the "prettier" kind, then this could stay singular.

analyzer_plugin/lib/src/model.dart
```diff
@@ -309,9 +309,13 @@ class InputElement extends AngularElementImpl {
   /// names of a input and the setter are the same.
   final SourceRange setterRange;
 
+  final List<String> alternativeNames;
```
MichaelRFairhurst: I'm wondering if this should be a list, or just `String alternativeName`. Do you see a case for having multiple coming up?

If multiple don't come up, a `final String` is safer than a `final List<String>` since the latter isn't checked against later additions/deletions (though I really hope we never even accidentally commit code that mutates these alternatives!).

Also, could use a comment for how this happens.

analyzer_plugin/lib/src/model.dart
```diff
@@ -309,9 +309,13 @@ class InputElement extends AngularElementImpl {
...
+
   InputElement(String name, int nameOffset, int nameLength, Source source,
-      this.setter, this.setterRange, this.setterType)
-      : super(name, nameOffset, nameLength, source);
+      this.setter, this.setterRange, this.setterType,
+      {List<String> alternativeNames})
```
mk13: That's right, I forgot.

analyzer_plugin/lib/src/model.dart
mk13: Hmm I was actually contemplating whether to use List or just a single String. I went with the safer approach for now, but I might revert it back to a single string case.

server_plugin/lib/src/completion.dart
```diff
@@ -414,8 +415,7 @@ class TemplateCompleter {
       suggestHtmlTags(template, suggestions);
       suggestTransclusions(target.parent, suggestions);
     }
-
-    return suggestions;
+    return new Set.from(suggestions).toList();
```
mk13: In the case of `[foo]` appearing in `OneFooDirective` and `OtherFooDirective`, I think it should be fine to suggest a single `[foo]` since from a completion perspective, it really doesn't care about the origin; we're not revealing the location of the completion in the completion drop-down. Only the resolver does (and future navigation). 

But that's a fair point, I can move the deduping into our input-list generation on the directives list instead of at the suggestion level.

server_plugin/lib/src/completion.dart
mk13: I think this is more of a design choice really - but for now I can suggestion the alternative name only if it exists.

analyzer_plugin/test/resolver_test.dart
```diff
@@ -378,6 +378,43 @@ class TestPanel {
...
+}
+''');
+    final code = r"""
+<span [class]='text'></span>
+""";
+    await angularDriver.getStandardHtml();
+    _addHtmlSource(code);
+    await _resolveSingleTemplate(dartSource);
+    errorListener.assertNoErrors();
+  }
```
mk13: Will add.

server_plugin/lib/src/completion.dart
MichaelRFairhurst: It's tricky because as long as those selectors are defined in different files, their `Location` info won't match, so `new Set.from` won't dedupe them.

As to whether or not we want to dedupe them, I guess that comes down to whether the editor uses the `Location` info or not. I'm inclined to say, since we never know that, we should assume it is being used, and provide both suggestions so that their `Location`s can be distinguished.

mk13: I think we should have a separate deduping strategy for each type. 

For Input types, since inputs have to be unique, `Set` style deduping on the directives should suffice. For attribute-directive deduping, we can use a name deduping scheme.

mk13:
Changed to single String instead of list on 'alternativeName'. Unique InputElement for standardHtmlAttribute.attributes is now generated once and stored instead of it being generated every time 'computeSuggestions' is called.

analyzer_plugin/lib/src/model.dart
```diff
@@ -309,12 +309,15 @@ class InputElement extends AngularElementImpl {
...
+  /// A given input can have an alternative name, or more 'conventional' name
+  /// that differs from the name provided by dart:html source.
+  /// For example: source -> 'className', but prefer 'class'.
+  final alternativeName;
 
   InputElement(String name, int nameOffset, int nameLength, Source source,
       this.setter, this.setterRange, this.setterType,
-      {List<String> alternativeNames})
-      : alternativeNames = alternativeNames ?? <String>[],
+      {alternativeName})
```
MichaelRFairhurst: this can be just `this.alternativeName` here

analyzer_plugin/lib/src/standard_components.dart
```diff
@@ -11,7 +11,13 @@ class StandardHtml {
   final Map<String, OutputElement> events;
   final Map<String, InputElement> attributes;
 
-  StandardHtml(this.components, this.events, this.attributes);
+  /// In attributes, there can be multiple strings that point to the
+  /// same [InputElement] generated from [alternativeInputs] (below).
+  /// This will provide a static source of unique [InputElement]s.
+  final List<InputElement> uniqueAttributeElements;
```
MichaelRFairhurst: This should probably stay a `Set` (they are more performant, so preferred unless you need to call a method only on list such as `[2]` or `sort()`)

also ohhhhhhh I was totally not following how this deduplication is required here, or you get `[class]` suggested twice, from the same exact object. Really like your placement, though, with this being cached :+1:

server_plugin/lib/src/completion.dart
```diff
@@ -742,40 +726,27 @@ class TemplateCompleter {
...
+    final completion = inputElement.alternativeName ?? inputElement.name;
+    return new CompletionSuggestion(CompletionSuggestionKind.INVOCATION,
+        defaultRelevance, completion, completion.length, 0, false, false,
+        element: element);
   }
 
   protocol.Element _createPlainAttributeElement(
       InputElement inputElement, protocol.ElementKind kind) {
-    final name = inputElement.name;
+    final name = inputElement.alternativeName ?? inputElement.name;
```
MichaelRFairhurst: I hate to have even more nits to pick on what's a small reliable PR :( but would it be better to do `name` and `originalName`, so code like this can just call `name` again? Instead of `alternativeName` and `name`?

server_plugin/test/completion_contributor_test.dart
```diff
@@ -1474,6 +1516,27 @@ class OtherComp {
...
+    ''');
+
+    addTestSource('<div ^></div>');
+
+    await resolveSingleTemplate(dartSource);
+    await computeSuggestions();
+    expect(replacementOffset, completionOffset);
+    expect(replacementLength, 0);
+    assertSuggestSetter('class', relevance: DART_RELEVANCE_DEFAULT - 2);
+    assertNotSuggested('className');
```
MichaelRFairhurst: ooh! GREAT test case!

analyzer_plugin/lib/src/standard_components.dart
mk13: Sure I'll set it as Set. Indexing is not being done so should be fine.

analyzer_plugin/lib/src/model.dart
mk13: Right. Will change.

server_plugin/lib/src/completion.dart
mk13: Sure, this can be done. My original concern was that it might have side-effects in the resolver that uses .name to point to the original (ex: className instead of class), but it should be fine I think since the resolver directly uses the InputElement and not the name.

mk13:
Perfectly fine :)

PR #345, opened 2017-05-18
Fix #309 empty ngIf suggests brackets. Some refactoring of completion

### LINKED ISSUE #309: NgIf without a value reports "is not a string, requires [bracket] syntax"
This is a *very* misleading error, based on how:

```
<div ngIf>
```
becomes
```
<template ngIf><div></div></template>
```

which is indeed a string binding, though most angular users wouldn't know it. And...even I am not 100% sure that I have this circumstance right.

Once you add a value,

```
<div ngIf="x">
```
it becomes
```
<template [ngIf]="x"></template>
```

so the solution is *not* to add bracket syntax, but to add a value (which, well, adds bracket syntax.).

I'm going to assume that people will follow the error's advice and get themselves into trouble, so, giving this p1.

MichaelRFairhurst:
Looks like the angular template parser creates an empty binding, not a
text attr, for things like `*ngIf` and `*ngFor="let item of"`. Match
that.

Then most of the work was about suppressing errors, because
`*ngFor="let item of items"` technically results in an empty `[ngFor]`
binding which neither matches an input nor has an expression. Suppress
that error, without suppressing the error for `*ngIf` without a value,
by checking that the input matches. Other empty bindings like `of`, or
`trackBy` have new checks in `converter.dart`. A misspelled `trackBy`
will still be caught by the old process.

This also had implications for completion of star attrs, which also had
to dance around these virtual `[ngFor]` bindings. I decided to not just
do the minimum change to get this working, but to also refactor
`findTarget` to be part of `DartSnippetExtractor` so that we can
leverage the cleanliness of the visitor pattern over nasty `is` checks
and spaghetti logic. Not that it was that bad. Also renamed it to
`CompletionTargetExtractor`. It can find `target` as well as
`dartSnippet` in one go.


analyzer_plugin/lib/src/converter.dart
```diff
@@ -815,8 +815,17 @@ class EmbeddedDartParser {
...
             ExpressionBoundType.input));
       } else {
-        attributes.add(new TextAttribute.synthetic(
-            key, keyOffset, null, null, originalName, originalNameOffset, []));
+        // Check for empty `of` and `trackBy` bindings, but NOT empty `ngIf`!
+        // NgFor (and other star directives) often result in a harmless, empty
+        // `[ngFor]` as a first attr. Don't flag it unless it matches an input
+        // (like `ngIf` does), which is checked by [SingleScopeResolver].
+        if (attributes.isNotEmpty) {
+          errorReporter.reportErrorForOffset(AngularWarningCode.EMPTY_BINDING,
```
mk13: 'isNotEmpty' and 'EMPTY_BINDING' had me scratching my head for a bit, but the comments helped out.

analyzer_plugin/test/resolver_test.dart
```diff
@@ -2850,6 +2884,41 @@ class TestPanel {
...
+class TestPanel {
+  String text; // 1
+}
+''');
+    final code = r'''
+<span *ngIf=""></span>
+''';
+    _addHtmlSource(code);
+    await _resolveSingleTemplate(dartSource);
+    // TODO would be nice if this selected ""
```
mk13: Should be doable under '_convertTemplateAttribute' under `if (ast is ParsedStarAst)`. You can access `valueToken` which has a `leftQuote.offset` and `rightQuote.offset`.

server_plugin/lib/src/completion.dart
```diff
@@ -33,53 +33,60 @@ import 'embedded_dart_completion_request.dart';
...
-          return target;
-        }
-      } else {
-        if (offsetContained(offset, child.openingNameSpan.offset,
-            child.openingNameSpan.length)) {
-          return child;
-        } else if (offsetContained(offset, child.offset, child.length)) {
-          return findTarget(offset, child);
-        }
+class CompletionTargetExtractor implements AngularAstVisitor {
```
mk13: Nice, a visitor approach to finding target 👍

analyzer_plugin/test/resolver_test.dart
MichaelRFairhurst: Hmm, unfortunately the error is only reported after its converted into `template="ngFor"` and then looked at in resolver.

Hmm, I suppose I could, in resolver, when I detect that `input.parent is TemplateElement`, report the error for `TemplateElement.valueRange` if its not null....

analyzer_plugin/lib/src/converter.dart
MichaelRFairhurst: Ooh that's a good point. I tried hard to write a clear comment...but I should specifically cover the not empty/ empty distinction, which I hadn't thought of.

analyzer_plugin/test/resolver_test.dart
mk13: Or better yet, we could just catch this at the AST level :)

zoechi:
> `*ngFor="let item of items"` technically results in an empty `[ngFor]`

I'm not sure about the context of this comment, but AFAIK this should result in
`ngFor [ngForOf]="items" let-item`

analyzer_plugin/test/resolver_test.dart
MichaelRFairhurst: Unfortunately we can't, because its legal to have an empty star binding. For instance, I think acx has a `*deferredContent` binding that takes no value. :(

MichaelRFairhurst:
@zoechi that's what i thought too. And then didn't think. And now think you're right.

The source here seems to tell the story that they produce the same thing, but simply with a null expression when the next character is let or EOF: https://github.com/dart-lang/angular2/blob/master/lib/src/compiler/expression_parser/parser.dart#L647

Aaaaand here is where the result is used: https://github.com/dart-lang/angular2/blob/ba179a1dac7c5e39bb5e29bc73874a02d34a9f25/lib/src/compiler/template_parser.dart#L258

I think we're still matching the visible behavior very well, though I would, all things being equal, rather have our implementation details match angular's. Will experiment with turning it back into text attrs and see if I can easily get all these tests passing with that solution.

MichaelRFairhurst:
Updated to handle @zoechi's feedback. Much better solution now, I think, though it took me some time to make sure the changes made for a *cleaner* solution

server_plugin/lib/src/completion.dart
```diff
@@ -488,10 +518,15 @@ class TemplateCompleter {
   ) async {
     final typeProvider = template.view.component.classElement.enclosingElement
         .enclosingElement.context.typeProvider;
-    final target = findTarget(request.offset, template.ast);
     final initialSuggestionsCount = collector.suggestionsLength;
     final replacementRangeCalculator = new ReplacementRangeCalculator(request);
-    target.accept(replacementRangeCalculator);
+    final extractor = new CompletionTargetExtractor(request.offset);
+    template.ast.accept(extractor);
+    final target = extractor.target..accept(replacementRangeCalculator);
```
mk13: Don't need double.

Issue #167, opened 2016-12-20
Templates should be able to use #, affect the parent scope


zoechi:
```html
<template #itemTemplate let-item="item" let-i="index">
  <list-item [label]="item.name" (click)="navigateToMenu(item)"></list-item>
</template>

<mdl-card *ngIf="projectService.project != null" class="mdl-shadow--2dp">
  <span card-title>Select</span>

  <div card-content>
    <list-component class="menus-list" [data]="projectService?.project?.menus"
                    [template]="itemTemplate" tabIndex="0"
                    [(ngModel)]="projectService.selectedMenu"
                    (ngModelChange)="navigateToMenu($event)"></list-component>

    <div class="buttons">
      <add-item-button (click)="addMenu()">Create a new menu</add-item-button>
      <remove-item-button (click)="removeMenu()"
                          [disabled]="projectService.selectedMenu == null">
        Delete the selected menu
      </remove-item-button>
    </div>
  </div>

</mdl-card>

<router-outlet></router-outlet>
```

results in 

```
ERROR: Undefined name 'itemTemplate'. ([fhir_client] lib/designer/webui/menu_editor/menus_editor_component.html:10)
```

for 

```html
 [template]="itemTemplate"
```

MichaelRFairhurst:
Hmmmm....this is a tricky one.

Usually, templates denote scopes. The issue is that the template is only in scope within the template.

Ah, I think I see where my confusion comes from: `let-item` makes a variable within the scope of `<template>`, while `#itemTemplate` makes a variable within the scope outside `<template>`? @matanlurey  can you confirm?

MichaelRFairhurst:
Ping @matanlurey  again

zoechi:
Friendly ping. 
This is still an issue.

MichaelRFairhurst:
Thanks for filing #321, that clarified this to me.

So, in a <template> tag, `#foo` and `let foo` are different, and `#foo` should affect the parent scope while `let` affects the child scope.

zoechi:
Exactly, I added some explanation in #321 that should make it more clear.

MichaelRFairhurst:
Awesome, that will definitely help!

mk13:
Taking a look at this, will have a fix soon.

mk13:
Should be fixed with PR #348 . Reopen if problem persists.

Issue #321, opened 2017-05-15
Wrong interpretation of `let-foo` on `<template>`


zoechi:
![image](https://cloud.githubusercontent.com/assets/405837/26053824/994522ae-396a-11e7-98a6-891ebd0a8a07.png)

The error message indicates that `let-item` is interpreted like what `#item` is supposed to do - to create a reference to the `<template>` element, but `let-item` is supposed to provide a reference to the `$implicit` context property passed to `createEmbeddedView` (where the component is stamped), like `item` in `*ngFor="let item of items"`.

Somewhat related to 

- #167
- #176

MichaelRFairhurst:
Hey again Gunter!

Interesting. I thought this was actually how it worked, I thought that html references and components were provided in `$implicit` and therefore bound via `#foo`.

You can see my source of confusion in the tests: [#x test](https://github.com/dart-lang/angular2/blob/9eb1ef80eb09ac72c0e24e0ea3a6dcd46d036544/test/compiler/expression_parser/parser_test.dart#L396) and [let x test](https://github.com/dart-lang/angular2/blob/9eb1ef80eb09ac72c0e24e0ea3a6dcd46d036544/test/compiler/expression_parser/parser_test.dart#L403), but this is only for inside a template binding. Outside of a template binding `#foo` is a `ReferenceAst`, and its handled completely differently.

This should be fixed: we should stop treating all references as `$implicit`, and not treat let bound vs hash bound the same way.

zoechi:
```html
<template #foo let-item let-bar="bar">
  <div>{{item}}</div>
  <div>{{bar}}</div>
</template>
<div [myDirective] [templateRef]="foo"></div>
```
```dart
@Directive(selector: '[myDirective]')
class MyDirective {
  final ViewContainerRef _viewContainerRef;
  NgTemplateOutlet(this._viewContainerRef);

  @Input()
  TemplateRef templateRef

  stampTemplate() {
    _insertedViewRef = _viewContainerRef.createEmbeddedView(templateRef);
    _insertedViewRef.setLocal(r'$implicit', 'implicitContent');
    _insertedViewRef.setLocal('bar', 'barContent');
  }
}
```
will result in 
```html
<div>implicitContent</div>
<div>bar</div>
```
- `#foo` introduces a template variable that reference the template element
- `let-item` introduces a context variable inside the template that references the item added with key `$implicit` to the context 
[...]

Issue #176, opened 2016-12-20
Free standing templates which declare variables report false errors and maybe can't be typechecked in general


zoechi:
```html
<template #itemTemplate let-item="item" let-i="index"></template>
```
This is used to pass the template to `[ngTemplateOutlet]` or `ngForTemplate` and the context passed there can be made available by declaring template variables like `let-item="item"` or just `let-context` for the `$implicit` property.

MichaelRFairhurst:
Hmm, where does the value for item come from? Does it come from how `#itemTemplate` is used, or from a previous `#item` declaration? Or is it a member of the enclosing class, defined in an outer `ngFor`?

If it depends on how `itemTemplate` is used, we may have to simply skip types of analyses on the contents. Or it may be unanalyzable, like ngFor would be if we didn't actually own it, it which case a means to opt out of the errors is the right solution.

zoechi:
```html
<template [ngTemplateOutlet]="templateRef" [ngOutletContext]="{'$implicit': {'item': 'a', 'index': 2}, 'other': 'otherValue'}"></template>
```
As mentioned a template can be used this way also with `ngForTemplate` and a context being passed that is made available to the template by declaring variables like `let-self` would refer to the `$implicit` property and `let-x="other"` to the `other` property of the context (like `let-item="item"` in my initial post).

where `templateRef` is a reference from 
```dart
@ContentChildren('itemTemplate') TemplateRef templateRef;`
```
A working example (TS) would be https://plnkr.co/edit/JuwbvVe5sBGMI9Xm2DsS?p=preview

Hope this helps.

MichaelRFairhurst:
Trying to think of a way to summarize what types of validation we could do here.

One option is to simply hide the error you reported originally. The variables should still be bound, but as a `dynamic`, which is correct because that's the most we know but we can still guarantee things around them.

A more complicated option is to treat templates as functions. We could validate them according to arguments that are applied to them, which means standard angular components using them such as ngFor and ngIf and deferred need to "call" the templates during analysis. Of course, the function overall can still be validated by the first option I proposed, as a minimum if its never "called."

How to detect when a template is "called" can involve multiple things as well. We probably have to hardcode how the core directives such as ngFor and ngIf call the template.

Angulardart doesn't yet support `ngOutletContext`, but that would be a fairly straightforward way to track a "called" and typecheck the "parameters".

More than that is probably more effort than worth it.

[...]

PR #348, opened 2017-05-19
Fix reference vs. let logic

mk13:
Pretty large change in terms of how let- and # variables are defined and resolved as.
Resolves issues: #167  #176  #321 

Major changes:
- `let-` binding is now only allowed to be used in `<template>` bindings, throws an error otherwise.
- `#ref` now properly references `TemplateRef` object when used inside template: `<template #ref>`
- `$implicit` logic is changed. Only in the case of `ngFor` will `$implicit` be defined. Otherwise, all other `let-noValue` bindings will link to a dynamic type. In addition, let bindings with value will also automatically be defined to dynamic type. (we can possibly, in the future, resolve these types)
- Added checks for ambiguous exportAs referencing. If two directives have the same `exportAs` value and both directives are used in the same tag linked to a `#ref`, it will throw an error. When this occurs in Angular, no error is thrown but that segment silently fails. Note that this will ONLY be thrown if ambigious exportAs value is attempted to be assigned to a reference, otherwise it won't complain. Example, suppose dir1 and dir2 have same `exportAs` value: `<div dir1 dir2></div>` won't throw an error. `<div dir1 dir2 #ref='ambigious'></div>` will complain.

Minor changes:
- In `_convertAttributes`, changed `<ParsedX>` types to be standard base type. Was doing bad forced casting. Better follows OOP design.

### DIFF
--- a/analyzer_plugin/lib/src/converter.dart
+++ b/analyzer_plugin/lib/src/converter.dart
@@ -172,6 +172,7 @@ class HtmlTreeConverter {
         events: node.events,
         properties: node.properties,
         references: node.references,
+        isTemplate: true,
       );
       final closeComponent = node.closeComplement;
       SourceRange openingSpan;
@@ -247,32 +248,44 @@ class HtmlTreeConverter {
   }
 
   List<AttributeInfo> _convertAttributes({
-    List<ParsedAttributeAst> attributes: const [],
-    List<ParsedBananaAst> bananas: const [],
-    List<ParsedEventAst> events: const [],
-    List<ParsedPropertyAst> properties: const [],
-    List<ParsedReferenceAst> references: const [],
-    List<ParsedStarAst> stars: const [],
+    List<AttributeAst> attributes: const [],
+    List<BananaAst> bananas: const [],
+    List<EventAst> events: const [],
+    List<PropertyAst> properties: const [],
+    List<ReferenceAst> references: const [],
+    List<StarAst> stars: const [],
+    bool isTemplate: false,
   }) {
     final returnAttributes = <AttributeInfo>[];
 
     for (final attribute in attributes) {
-      if (attribute.name == 'template') {
-        returnAttributes.add(_convertTemplateAttribute(attribute));
-      } else {
-        String value;
-        int valueOffset;
-        if (attribute.valueToken != null) {
-          value = attribute.valueToken.innerValue.lexeme;
-          valueOffset = attribute.valueToken.innerValue.offset;
+      if (attribute is ParsedAttributeAst) {
+        if (attribute.name == 'template') {
+          returnAttributes.add(_convertTemplateAttribute(attribute));
+        } else if (attribute.name.startsWith('let-') && !isTemplate) {
+          final offset = attribute.nameOffset;
+          final length = attribute.name.length;
+          errorListener.onError(new AnalysisError(
+            templateSource,
+            offset,
+            length,
+            AngularWarningCode.INVALID_LET_BINDING,
+          ));
+        } else {
+          String value;
+          int valueOffset;
+          if (attribute.valueToken != null) {
+            value = attribute.valueToken.innerValue.lexeme;
+            valueOffset = attribute.valueToken.innerValue.offset;
+          }
+          returnAttributes.add(new TextAttribute(
+            attribute.name,
+            attribute.nameOffset,
+            value,
+            valueOffset,
+            dartParser.findMustaches(value, valueOffset),
+          ));
         }
-        returnAttributes.add(new TextAttribute(
-          attribute.name,
-          attribute.nameOffset,
-          value,
-          valueOffset,
-          dartParser.findMustaches(value, valueOffset),
-        ));
       }
     }
     bananas.map(_convertExpressionBoundAttribute).forEach(returnAttributes.add);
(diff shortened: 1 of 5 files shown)


analyzer_plugin/lib/src/converter.dart
```diff
@@ -247,32 +248,44 @@ class HtmlTreeConverter {
...
+        if (attribute.name == 'template') {
+          returnAttributes.add(_convertTemplateAttribute(attribute));
+        } else if (attribute.name.startsWith('let-') && !isTemplate) {
+          final offset = attribute.nameOffset;
+          final length = attribute.name.length;
+          errorListener.onError(new AnalysisError(
+            templateSource,
+            offset,
+            length,
+            AngularWarningCode.INVALID_LET_BINDING,
```
MichaelRFairhurst: should this be caught in angular_ast?

analyzer_plugin/lib/src/converter.dart
```diff
@@ -247,32 +248,44 @@ class HtmlTreeConverter {
...
+          ));
+        } else {
+          String value;
+          int valueOffset;
+          if (attribute.valueToken != null) {
+            value = attribute.valueToken.innerValue.lexeme;
+            valueOffset = attribute.valueToken.innerValue.offset;
+          }
+          returnAttributes.add(new TextAttribute(
```
MichaelRFairhurst: we could make this a `LetAttribute`, new to our ast. Would that be difficult do you think?

analyzer_plugin/lib/src/converter.dart
```diff
@@ -247,32 +248,44 @@ class HtmlTreeConverter {
...
     for (final attribute in attributes) {
-      if (attribute.name == 'template') {
-        returnAttributes.add(_convertTemplateAttribute(attribute));
-      } else {
-        String value;
-        int valueOffset;
-        if (attribute.valueToken != null) {
-          value = attribute.valueToken.innerValue.lexeme;
-          valueOffset = attribute.valueToken.innerValue.offset;
+      if (attribute is ParsedAttributeAst) {
```
MichaelRFairhurst: should the synthetics just get dropped?

analyzer_plugin/lib/src/converter.dart
```diff
@@ -382,7 +397,7 @@ class HtmlTreeConverter {
     final value = ast.value;
     if ((value == null || value.isEmpty) &&
         !ast.prefixToken.errorSynthetic &&
-        !ast.suffixToken.errorSynthetic) {
+        (!ast?.suffixToken?.errorSynthetic ?? true)) {
```
MichaelRFairhurst: should `ast` have a `?.` we know the ast isn't null, right?

Also, order is a bit unclear. I'm pretty sure this is equivalent to `(!(a?.b?.c ?? true))`, but it might be clearer as `!(a?.b?.c ?? false)`

analyzer_plugin/lib/src/converter.dart
```diff
@@ -447,8 +462,9 @@ class HtmlTreeConverter {
 
     final value = parsed.valueToken?.innerValue?.lexeme;
     if ((value == null || value.isEmpty) &&
-        !parsed.prefixToken.errorSynthetic &&
-        !parsed.suffixToken.errorSynthetic) {
+            !parsed.prefixToken.errorSynthetic &&
+            !parsed?.suffixToken?.errorSynthetic ??
```
MichaelRFairhurst: same comment regarding order of operations and `parsed?.` instead of `parsed.`

analyzer_plugin/lib/src/resolver.dart
```diff
@@ -442,80 +441,146 @@ class PrepareScopeVisitor extends AngularScopeVisitor {
...
+  Map<String, List<InternalVariable>> _defineExportAsVariables(
+      List<AbstractDirective> directives) {
+    final exportAsMap = <String, List<InternalVariable>>{};
+    for (final directive in directives) {
       final exportAs = directive.exportAs;
       if (exportAs != null) {
         final name = exportAs.name;
         final type = directive.classElement.type;
-        internalVariables[name] = new InternalVariable(name, exportAs, type);
+        exportAsMap.putIfAbsent(name, () => <InternalVariable>[]);
```
MichaelRFairhurst: nice on catching the conflicts!

analyzer_plugin/test/abstract_angular.dart
```diff
@@ -76,7 +76,8 @@ ResolvedRange getResolvedRangeAtString(
     }
     return false;
   }, orElse: () {
-    fail('ResolvedRange at $offset was not found in [\n${ranges.join('\n')}]');
+    fail(
+        'ResolvedRange at $offset of $str was not found in [\n${ranges.join('\n')}]');
```
MichaelRFairhurst: ooh good addition here too

analyzer_plugin/lib/src/converter.dart
mk13: I think we should preserve synthetics at one point for quickfix. Would be nice to just scan on a single token and figure out what is synthetic within it.

analyzer_plugin/lib/src/converter.dart
mk13: I think that's a good approach, it fits the ast-model needs. I'll keep this for now and move the code later.

analyzer_plugin/lib/src/converter.dart
mk13: Not really! I could implement into the ast and merge these changes along with the `let-` binding error.

analyzer_plugin/lib/src/converter.dart
mk13: I've simply added it to keep the IDE to stop from complaining, but logistically speaking - you are correct. I will see if I can find a more elegant solution around this entire subblock.

mk13: Fix it to be more elegant.

mk13:
Next steps is to move `let-` binding into the AST-level instead of here and remove the redundant error notifications.

PR #350, opened 2017-05-22
Remove redundant let error

mk13:
Code will be broken temporarily until this is merged.

### DIFF
--- a/analyzer_plugin/lib/ast.dart
+++ b/analyzer_plugin/lib/ast.dart
@@ -195,6 +195,8 @@ class TextAttribute extends AttributeInfo {
 }
 
 class Mustache extends AngularAstNode {
+  static const EMPTY_LIST = const <Mustache>[];
+
   Expression expression;
   @override
   final int offset;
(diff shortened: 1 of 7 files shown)


analyzer_plugin/lib/ast.dart
```diff
@@ -195,6 +195,8 @@ class TextAttribute extends AttributeInfo {
 }
 
 class Mustache extends AngularAstNode {
+  static const EMPTY_LIST = const <Mustache>[];
```
liuming0: use camelCase.

mk13: There's a bit of SCREAMING_CAPS already happening within the code, so I'll just leave this to stay consistent. (https://www.dartlang.org/guides/language/effective-dart/style#prefer-using-lowercamelcase-for-constant-names)

scheglov: FWIW, in Analyzer we at the end decided to use explicit `const <TypeName>[]` - it is not longer and not more expensive. We didn't migrate, but try to write new code in this style.

mk13: Oh sure, good to know. I'll go ahead and add those changes into this PR as well then.

PR #355, opened 2017-05-24
Autocompletion plugin

mk13:
Enables autocompletion support via plugin. Will not work until a submitted chromium change 2900383002 is submitted.

### DIFF
--- a/analyzer_plugin/lib/starter.dart
+++ b/analyzer_plugin/lib/starter.dart
@@ -26,7 +26,8 @@ class Starter {
     ContextBuilder.onCreateAnalysisDriver = onCreateAnalysisDriver;
     server
       ..onResultErrorSupplementor = sumErrors
-      ..onNoAnalysisResult = sendHtmlResult;
+      ..onNoAnalysisResult = sendHtmlResult
+      ..onNoAnalysisCompletion = sendAngularCompletions;
   }
 
   void onCreateAnalysisDriver(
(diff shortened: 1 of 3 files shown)


analyzer_plugin/lib/starter.dart
```diff
@@ -86,51 +85,6 @@ class Starter {
 
     sendFn(null, null, null);
   }
-
```
MichaelRFairhurst: can we leave this file as it was before? So that we can use this Starter for google3 users and the new_plugin for external, without having to branch?

new_plugin/lib/plugin.dart
```diff
@@ -120,25 +121,27 @@ class AngularAnalysisPlugin extends ServerPlugin {
...
+      throw new plugin.RequestFailure(plugin.RequestErrorFactory
+          .pluginError('Failed to analyze $filePath', null));
+    }
+    final AngularDriver driver = driverMap[contextRoot];
+    final analysisResult = await driver.dartDriver.getResult(filePath);
+    final contributor = new AngularCompletionContributor(driver);
+    final performance = new CompletionPerformance();
+    final fileSource = resourceProvider.getFile(filePath).createSource();
+    final request = new CompletionRequestImpl(analysisResult, resourceProvider,
+        fileSource, parameters.offset, performance, null);
```
MichaelRFairhurst: What's the last null item? Could use a comment explaining what it is and why its not needed, if it isn't.

analyzer_plugin/lib/starter.dart
mk13: That should be possible, I'll have to revert some of the changes in the analysis_server side then.

MichaelRFairhurst: Oh shoot. Still, I think its worth the effort, otherwise we won't be able to dogfood this inside google until the whole new arch is ready.

Issue #334, opened 2017-05-17
suggest attributes from selectors in loaded directives


MichaelRFairhurst:
Note that they should be lower priority, as its more common that `[foo]` is matching an input (ie `[foo]`) than an actual attr.

Selectors with values (ie `[foo=bar]`) should suggest both the half and whole tamale, ie `foo` and `foo="bar"`, I would think.

MichaelRFairhurst:
Handled in #358

Issue #333, opened 2017-05-17
Suggest inputs/outputs for unmatched directives


MichaelRFairhurst:
If I have a directive like

```
@Directive(selector: '[foo]', inputs: ['foo'])
class Foo { ... }
```

I will never get recommended `[foo]`. Why? Because we only recommend inputs for matched directives. `Foo` isn't matched until we have `[foo]`, and once we have `[foo]` we won't suggest it again.

Most relevant example: `NgModel`. (kinda a big deal)

Of course, `Foo` may have another input, and an output, which don't correlate with the selector. These won't be suggested until '[foo]' has been bound, which may not match the most readable API. For instance, perhaps `<div (trigger)="bar" [onUniversal]="click">` where `[onUniversal]` matches some directive with output `trigger`. Not sure how common this is.

We could do one or both of
* suggest all directive inputs regardless of whether they've been matched yet, at a very low priority
* suggest self-fulfilling inputs like these, of unmatched directives, at a lowered priority

note that we shouldn't need to handle this problem for Components, because the vast majority of components are matched correctly at the correct time since they use tag names. Of course, this is not guaranteed, either.

I say we start with the second approach, for directives only, and wait for feedback that inputs are missing.

mk13:
Done.

Issue #233, opened 2017-01-19
Can put ngModel on a li?


MichaelRFairhurst:
This passes our analyzer, and, almost certainly shouldn't:

```
  <li [(ngModel)]="x">...<li>
```

Seems like anything can have `ngModel`.

As a short term solution, we could make a whitelist (or even a blacklist) of elements where an `ngModel` can/cannot go.

I have a feeling that the reason why `ngModel` can't go on a `li` is that `ngModel` uses like `@HostAttribute` or something. In that case, rather than use a whitelist/blacklist, we can check that the attribute its used on has the targeted attribute.

This is actually a pretty big deal, would make a big case for validating these next-level annotations in angular, which I believe are in the cheatsheet, as unsupported.

zoechi:
`ngModel` can go on any element that is either a native input element or where a component or directive is applied that implements `ControlValueAccessor`

zoechi:
See also https://github.com/dart-lang/angular2/blob/1b71b61f73f43adcede21470c62e943bed635f69/lib/src/common/forms/directives/ng_model.dart#L76 where `NG_VALUE_ACCESSOR` is the only non-optional dependency

`ControlValueAccessor`s are registered with `NG_VALUE_ACCESSOR` (see https://github.com/dart-lang/angular2/blob/164b20a4cd06a1953efcc3ce15d302b31557570e/lib/src/common/forms/directives/default_value_accessor.dart#L8 for an example)

PR #357, opened 2017-05-29
Twoway autocomplete

mk13:
Basic two-way autocompletion added.

Currently does not support imported two-way bindings such as [(ngModel)]; this will be submitted in a separate PR as more additional work on directives will be needed to support this.

Since bananas are a weird hybrid of Setter(Input) and Getter(Output), bananas are Setters with a returnType (which are standard in Getters).

### DIFF
--- a/analyzer_plugin/lib/starter.dart
+++ b/analyzer_plugin/lib/starter.dart
@@ -26,8 +26,8 @@ class Starter {
     ContextBuilder.onCreateAnalysisDriver = onCreateAnalysisDriver;
     server
       ..onResultErrorSupplementor = sumErrors
-      ..onNoAnalysisResult = sendHtmlResult;
-    //..onNoAnalysisCompletion = sendAngularCompletions;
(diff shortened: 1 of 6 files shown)


MichaelRFairhurst:
> this will be submitted in a separate PR as more additional work on directives will be needed to support this.

Do you mean like #333? I'm still not sure if we want to do that, or #334. Or do you mean, factor in whatever validation logic we put in place for #233?

mk13:
Yup PR #333 will be needed to allow for NgModel. I'm not sure why we don't want to suggest [(ngModel)] through potential directives? I'm currently working on a solution to go through all in-scope (but not bound directives) and doing a loose "potential fit" scan on the current ElementInfo for any hard violations. 

For example: 
```
@Directive(
  selector: '[someDirectiveAttribute]:not(div)'
)...
```

If attempted to be used on `<div ^></div>` - this won't be suggested.

PR #358, opened 2017-06-02
Directives autocomplete

mk13:
Added support for suggesting plain attributes, inputs, outputs, and bananas from 'availableDirectives'.

`availableDirective`: Given a directive that is not bounded to an Element, it is considered available to that Element if and only if there are no violations with its selector. One notable quirk is that an attribute-selector will always be available to the Element. However, a not:attribute-selector will check the Element to see if the not:attribute exists in the Element.

During suggestion phase: 
* For a directive in `availableDirectives`, extract all attribute-selectors.
* Given the list of attribute-selectors, if there is an Input with the same name - drop the plain attribute form. 
* If there is an input that overrides AND there is a corresponding `inputNameChange` output, suggest all three: input, output, and banana syntax.

Resolves issues #333 and #328 (partially).

### DIFF
--- a/analyzer_plugin/lib/src/selector.dart
+++ b/analyzer_plugin/lib/src/selector.dart
@@ -232,7 +232,7 @@ class ClassSelector extends Selector {
     return SelectorMatch.NonTagMatch;
   }
 
-  // Returns true if the element has a 'class' attribute.
+  // Returns true if the element matches the '.<className>'.
   @override
   bool availableTo(ElementView element) =>
       match(element, null) == SelectorMatch.NonTagMatch;
--- a/analyzer_plugin/lib/src/selector.dart
+++ b/analyzer_plugin/lib/src/selector.dart
@@ -10,7 +10,7 @@ import 'package:meta/meta.dart';
 enum SelectorMatch { NoMatch, NonTagMatch, TagMatch }
 
 /// The [Selector] that matches all of the given [selectors].
-class AndSelector extends Selector {
+class AndSelector extends Selector implements CompoundSelector {
   final List<Selector> selectors;
 
   AndSelector(this.selectors);
@@ -45,6 +45,20 @@ class AndSelector extends Selector {
       }
     }
     return true;
+  }
+
+  @override
+  List<AttributeSelector> get getAttributeSelectors {
+    final attributeSelectors = <AttributeSelector>[];
+    for (final selector in selectors) {
+      if (selector is AttributeSelector) {
+        attributeSelectors.add(selector);
+      } else if (selector is CompoundSelector) {
+        attributeSelectors
+            .addAll((selector as CompoundSelector).getAttributeSelectors);
+      }
+    }
+    return attributeSelectors;
   }
 
   @override
@@ -119,6 +133,7 @@ class AttributeSelector extends Selector {
     return SelectorMatch.NonTagMatch;
   }
 
+  // Want to always return true since this doesn't narrow scope.
   @override
   bool availableTo(ElementView element) => true;
 
@@ -166,7 +181,8 @@ class AttributeValueRegexSelector extends Selector {
   }
 
   @override
-  bool availableTo(ElementView element) => true;
+  bool availableTo(ElementView element) =>
+      match(element, null) == SelectorMatch.NonTagMatch;
 
   @override
   String toString() => '[*=$regexpStr]';
@@ -212,22 +228,14 @@ class ClassSelector extends Selector {
     // add resolved range
     final valueOffset = element.attributeValueSpans['class'].offset;
     final offset = valueOffset + index;
-    template.addRange(new SourceRange(offset, name.length), nameElement);
+    template?.addRange(new SourceRange(offset, name.length), nameElement);
     return SelectorMatch.NonTagMatch;
   }
 
+  // Returns true if the element has a 'class' attribute.
   @override
-  bool availableTo(ElementView element) {
-    final name = nameElement.name;
-    final val = element.attributes['class'];
-    if (val == null) {
-      return false;
-    }
-    if (!val.split(' ').contains(name)) {
-      return false;
-    }
-    return true;
-  }
+  bool availableTo(ElementView element) =>
+      match(element, null) == SelectorMatch.NonTagMatch;
 
   @override
   String toString() => '.${nameElement.name}';
@@ -275,10 +283,8 @@ class ElementNameSelector extends Selector {
   }
 
   @override
-  bool availableTo(ElementView element) {
-    final name = nameElement.name;
-    return element.localName == name;
-  }
+  bool availableTo(ElementView element) =>
+      nameElement.name == element.localName;
 
   @override
   String toString() => nameElement.name;
@@ -310,7 +316,7 @@ abstract class ElementView {
 }
 
 /// The [Selector] that matches one of the given [selectors].
-class OrSelector extends Selector {
+class OrSelector extends Selector implements CompoundSelector {
   final List<Selector> selectors;
 
   OrSelector(this.selectors);
@@ -337,6 +343,20 @@ class OrSelector extends Selector {
       }
     }
     return false;
+  }
+
+  @override
+  List<AttributeSelector> get getAttributeSelectors {
+    final attributeSelectors = <AttributeSelector>[];
+    for (final selector in selectors) {
+      if (selector is AttributeSelector) {
+        attributeSelectors.add(selector);
+      } else if (selector is CompoundSelector) {
+        attributeSelectors
+            .addAll((selector as CompoundSelector).getAttributeSelectors);
+      }
+    }
+    return attributeSelectors;
   }
 
   @override
@@ -374,7 +394,14 @@ class NotSelector extends Selector {
           : SelectorMatch.NoMatch;
 
   @override
-  bool availableTo(ElementView element) => !condition.availableTo(element);
+  bool availableTo(ElementView element) {
+    if (condition is AttributeSelector) {
+      // Need separate logic since AttributeSelector.availableTo
+      // always returns true.
+      return condition.match(element, null) == SelectorMatch.NoMatch;
+    }
+    return !condition.availableTo(element);
+  }
 
   @override
   String toString() => ":not($condition)";
@@ -438,7 +465,8 @@ abstract class Selector {
 
   /// Check whether the given [element] can potentially match with
   /// this selector. Or simply put, if there is no violation
-  /// then the given [element] is 'availableTo' this selector.
+  /// then the given [element] is 'availableTo' this selector without
+  /// contradiction.
   bool availableTo(ElementView element);
 
   /// See [HtmlTagForSelector] for info on what this does.
@@ -455,6 +483,10 @@ abstract class Selector {
   }
 
   void recordElementNameSelectors(List<ElementNameSelector> recordingList);
+}
+
+abstract class CompoundSelector {
+  List<AttributeSelector> get getAttributeSelectors;
 }
 
 enum _SelectorRegexMatch {
--- a/analyzer_plugin/lib/src/selector.dart
+++ b/analyzer_plugin/lib/src/selector.dart
@@ -10,7 +10,7 @@ import 'package:meta/meta.dart';
 enum SelectorMatch { NoMatch, NonTagMatch, TagMatch }
 
 /// The [Selector] that matches all of the given [selectors].
-class AndSelector extends Selector implements CompoundSelector {
+class AndSelector extends Selector {
   final List<Selector> selectors;
 
   AndSelector(this.selectors);
@@ -38,28 +38,14 @@ class AndSelector extends Selector implements CompoundSelector {
   }
 
   @override
-  bool availableTo(ElementView element) {
-    for (final selector in selectors) {
-      if (!selector.availableTo(element)) {
-        return false;
-      }
-    }
-    return true;
-  }
+  bool availableTo(ElementView element) =>
+      selectors.every((selector) => selector.availableTo(element));
 
   @override
-  List<AttributeSelector> get getAttributeSelectors {
-    final attributeSelectors = <AttributeSelector>[];
-    for (final selector in selectors) {
-      if (selector is AttributeSelector) {
-        attributeSelectors.add(selector);
-      } else if (selector is CompoundSelector) {
-        attributeSelectors
-            .addAll((selector as CompoundSelector).getAttributeSelectors);
-      }
-    }
-    return attributeSelectors;
-  }
+  List<AttributeSelector> getAttributeSelectors(ElementView element) =>
+      selectors
+          .expand((selector) => selector.getAttributeSelectors(element))
+          .toList();
 
   @override
   String toString() => selectors.join(' && ');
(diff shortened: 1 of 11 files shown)


analyzer_plugin/lib/src/selector.dart
```diff
@@ -196,10 +228,15 @@ class ClassSelector extends Selector {
     // add resolved range
     final valueOffset = element.attributeValueSpans['class'].offset;
     final offset = valueOffset + index;
-    template.addRange(new SourceRange(offset, name.length), nameElement);
+    template?.addRange(new SourceRange(offset, name.length), nameElement);
     return SelectorMatch.NonTagMatch;
   }
 
+  // Returns true if the element has a 'class' attribute.
```
liuming0: Does this description match the content?

server_plugin/bin/server.dart
```diff
@@ -19,5 +19,5 @@ void main(List<String> args) {
   ];
   final server = starter.start(args);
 
-  new ng.Starter().start(server);
+  //new ng.Starter().start(server);
```
liuming0: Remove this line.

server_plugin/lib/src/completion.dart
```diff
@@ -696,6 +737,76 @@ class TemplateCompleter {
...
+  /// and inputs (if name overlaps with attribute-directive).
+  void suggestFromAvailableDirectives(
+    Set<AbstractDirective> availableDirectives,
+    List<CompletionSuggestion> suggestions, {
+    bool suggestInputs: false,
+    bool suggestOutputs: false,
+    bool suggestBananas: false,
+    bool suggestPlainAttributes: false,
+  }) {
+    for (final directive in availableDirectives) {
```
MichaelRFairhurst: once again, great edge case here on not suggesting for unavailable directives.

server_plugin/lib/src/completion.dart
```diff
@@ -696,6 +737,76 @@ class TemplateCompleter {
...
+      final selector = directive.selector;
+      final attributeSelectors = <String, AttributeSelector>{};
+      final validInputs = <InputElement>[];
+
+      if (selector is AttributeSelector) {
```
MichaelRFairhurst: once again, if you move `get attributeSelectors` to `Selector`, you won't have to do the is check here.

server_plugin/lib/src/completion.dart
```diff
@@ -696,6 +737,76 @@ class TemplateCompleter {
...
+            (output) => output.name == outputComplement,
+            orElse: () => null);
+        if (output != null) {
+          if (suggestBananas) {
+            suggestions.add(_createBananaSuggestion(
+                input,
+                DART_RELEVANCE_DEFAULT,
+                _createBananaElement(input, protocol.ElementKind.SETTER)));
+          }
+          if (suggestOutputs) {
```
MichaelRFairhurst: this is the wrong suggestion, no?

```
@Directive(selector="[universalClick]")
class UniversalClickDirective {
  @Output()
  Stream<UniversalClickEvent> universalClick;
}
```

won't suggest `(universalClick)`. And

```
@Directive(selector="[universalClick]") 
class UniversalClickDirective {
  @Output()
  Stream<UniversalClickEvent> universalClickChange;
}
```

will suggest `(universalClickChange)` even though that directive wouldn't be matched until you added the `universalClick` property ie `<div universalClick (universalClickChange)="bar">`

server_plugin/test/completion_contributor_test.dart
```diff
@@ -2074,6 +2074,293 @@ class OtherComp {
...
+
+    addTestSource('<my-tag ^></my-tag>');
+
+    await resolveSingleTemplate(dartSource);
+    await computeSuggestions();
+    expect(replacementOffset, completionOffset);
+    expect(replacementLength, 0);
+    assertNotSuggested('myDirective');
+    assertSuggestSetter('[myDirective]');
+    assertSuggestGetter('(myDirectiveChange)', 'String');
```
MichaelRFairhurst: I don't think this should be suggested, because `<my-tag (myDirectiveChange)="x">` would complain that `(myDirectiveChange)` is not bound.

server_plugin/test/completion_contributor_test.dart
```diff
@@ -2074,6 +2074,293 @@ class OtherComp {
...
+    ''');
+    addTestSource('<my-tag ^></my-tag>');
+
+    await resolveSingleTemplate(dartSource);
+    await computeSuggestions();
+    expect(replacementOffset, completionOffset);
+    expect(replacementLength, 0);
+    assertNotSuggested('myDirective');
+    assertSuggestSetter('[myDirective]');
+    assertSuggestGetter('(myDirectiveChange)', 'String');
```
MichaelRFairhurst: again

server_plugin/test/completion_contributor_test.dart
```diff
@@ -2074,6 +2074,293 @@ class OtherComp {
...
+    assertSuggestSetter('[(myDirective)]', returnType: 'String');
+    assertSuggestSetter('foo');
+    assertSuggestSetter('bar');
+  }
+
+  // ignore: non_constant_identifier_names
```
MichaelRFairhurst: new test case:

```
@Directive(selector: '[foo]')
class Foo {
  @Output()
  Stream<Bar> foo;
}
...
html:
<my-tag ^> // assert suggest (foo)
```

analyzer_plugin/lib/src/selector.dart
```diff
@@ -38,6 +38,30 @@ class AndSelector extends Selector {
...
+    return true;
+  }
+
+  @override
+  List<AttributeSelector> get getAttributeSelectors {
+    final attributeSelectors = <AttributeSelector>[];
+    for (final selector in selectors) {
+      if (selector is AttributeSelector) {
+        attributeSelectors.add(selector);
+      } else if (selector is CompoundSelector) {
```
mk13: I think adding a visitor at this point might be over-kill. It's a lot of overhead for only serving one real purpose at the moment; should the same requirement arise but for another scenario, we can add this in. 

In the mean time, we can go with the first solution.

MichaelRFairhurst: I'm not sure if its the right time or not, but we do have an algorithm to find suggestable tags which is about the same; so the visitor could do that.

Of course, there are a number of algorithms we have on Selector: match, constrainSuggestions, even recording ranges could be separated from matching. Those could all potentially be visitors.

analyzer_plugin/lib/src/selector.dart
```diff
@@ -38,6 +38,30 @@ class AndSelector extends Selector {
...
+    for (final selector in selectors) {
+      if (!selector.availableTo(element)) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  @override
+  List<AttributeSelector> get getAttributeSelectors {
```
mk13: I think this would fall under "should produce the same thing unless explicit change" since an explicit change occurs when the file is re-analyzed; until the file is re-analyzed, it will always produce the same result. 

At the moment, I don't see any use-cases where we will be modifying this to receive parameters. So for now I'll keep it as a getter, but should the need arise later to no longer make it a getter, we can change it then.

analyzer_plugin/lib/src/selector.dart
```diff
@@ -153,6 +181,10 @@ class AttributeValueRegexSelector extends Selector {
   }
 
   @override
+  bool availableTo(ElementView element) =>
+      match(element, null) == SelectorMatch.NonTagMatch;
```
mk13: This one was a tough one to make a judgement call on.

I guess we can suggest `foo` if and only if `foo` does not already exist in the given element tag.

MichaelRFairhurst: > if and only if foo does not already exist in the given element tag.

I would add that it should be available to tags it matches. Otherwise `[foo=bar*][bang]` isn`t available to `<div foo="barbarian">` because the top level `AndSelector` will see that its first selector `[foo=bar*]` isn't available to the div since it has an attribute `foo`.

analyzer_plugin/lib/src/selector.dart
```diff
@@ -196,10 +228,15 @@ class ClassSelector extends Selector {
...
     final offset = valueOffset + index;
-    template.addRange(new SourceRange(offset, name.length), nameElement);
+    template?.addRange(new SourceRange(offset, name.length), nameElement);
     return SelectorMatch.NonTagMatch;
   }
 
+  // Returns true if the element matches the '.<className>'.
+  @override
+  bool availableTo(ElementView element) =>
+      match(element, null) == SelectorMatch.NonTagMatch;
```
mk13: This one should be restricted by existence of the class; primarily in cases of compound selectors.

`Component(selector: '[my-directive].foo')`

Then in `<div class="foo" ^>` should suggest `my-directive`.
On the other hand, `<div ^>` should not suggest `my-directive`.

MichaelRFairhurst:
Do we also want to handle the case of

```
@Directive(selector: '[foo][bar]')
class MyDirective {
  @Input()
  int bar;
}
```

where arguably `[bar]` shouldn't be suggested unless the element has a `foo` attr? Is there an easy solution for that?

in this case `foo` could be suggested, because its harmless, but `[bar]` would result in "unknown input bar" errors. 

On the other hand, 

```
@Directive(selector: '[foo][bar]')
class MyDirective {
  @Input()
  int foo;
  @Input()
  int bar;
}
```

this code has no error-free path from `<div>` to `<div [foo]="1" [bar]="2">`.....`<div [foo]="1">` lacks a `bar` attribute, `<div [bar]="1">` lacks a `foo` attribute, and '<div foo [bar]="2">`/`<div [foo]="1" bar>` have string bindings of int inputs (and yeah, it'd be ridiculous to expect people to temporarily add some properties to their elements just to get autocomplete to work)...

I guess we probably care more about suggesting '[foo]' and '[bar]' in the latter case than about the error someone might get for `[bar]` in the former case. And I certainly don't think we care about either case enough to try and *detect* these cycles.

Still, it poses and interesting question.

analyzer_plugin/lib/src/selector.dart
MichaelRFairhurst: Hmm, I agree on the final statement, but I don't think the name is right if that's what its doing.

Note that it for `[bar][baz]`, it will suggest both `bar` and `baz` even though it requires both to match. From that standpoint I'd think for `[my-directive].foo` that `my-directive` should be suggested even if it still also requires `class="foo"` to match.

If we were to start suggesting available classnames, the classnames must be considered available before they are used. 

The idea I get from the name `isAvailableTo` is that you can get to a selector match without having to remove/change anything -- additions only. From that standpoint `.bar` is available to `<div class="bar">`.

If we're calculating something else, we should have a different name.

Maybe what we want to do is count, rather than return a boolean? Like, "this directive is 1 class selector and 2 property selectors (one of which is an input) from matching this element." And then we could use that info to first suggest the classname, then suggest the non-input property, then suggest the input, because that path results in no errors along the way?

Alternatively, maybe there's a way to generalize this, like, `isAvailableToByAdding(ElementView view, {properties: true, classes: false, ...})`? Then we can keep the current behavior (which lets us avoid suggesting `my-directive` in your example) while still allowing us to suggest classnames in the future, and maybe a clearer API in terms of expressing its complexities?

analyzer_plugin/lib/src/selector.dart
mk13: Oh it will still yield true if `foo` does exist since it matches the criteria. 

In the case of `[foo=bar*][bang]`: 
* `<div foo="barbarian">` ; selector.availableTo => true since `[foo=bar*]` is true and `[bang]` is true.
* `<div foo="hello">`; selector.availableTo => false since `[foo=bar*]` is false and `[bang]` is true.

analyzer_plugin/lib/src/selector.dart
mk13: I'll probably move the mixin out and just put it in the abstract class `Selector` for now - then do another PR to refactor every Selector action/behavior into a visitor pattern

analyzer_plugin/lib/src/selector.dart
mk13: After mulling this over, I might change this to just a non-getter.

MichaelRFairhurst: I can still see it either way.

FWIW I got back a response to my request to expand effective dart's advice here, regarding toString() violating it. The response was that there's a line I missed: "If the name of the method starts with get or is a noun phrase like length or size that’s a sign you’re better off using a getter. You should define a getter instead of a method when....." Since toString isn't a noun phrase and doesn't start with get, it shouldn't be a getter, at least, not without renaming. And then I got a list of reasons why it shouldn't be renamed.

Granted, this all started with me asking why toString() wasn't a getter in the first place. And one response was "There's also an argument that toString() should be a method because it doesn't describe a property of the object so much as it transforms the object into a new one. It may be that's why we never changed it," which makes me think the part about "conceptually a property" is more of a judgement call than effective dart lets on.

So there's still two ways to look at it,

- to string only isn't a getter because of naming. This has getter naming and should be a getter
- to string isn't named like a getter because its not conceptually a property, which is very important for getters

I think the cleanest, most defensible, is the getter approach, and the safest given ambiguous advice is to make it a method.

Your call. I've already gotten overly involved in this simple choice out of curiousity. Never worked in a language with custom getters before, still learning how to best use them.

analyzer_plugin/lib/src/selector.dart
mk13: Just realized we've been discussing this under `AttributeValueRegexSelector` when this should be under `AttributeSelector`.

analyzer_plugin/lib/src/selector.dart
mk13: Going with the policy that "selector match without having to remove/change existing decorators" sounds like a good, uniform strategy.

server_plugin/lib/src/completion.dart
mk13: Was limiting scope to Input namespace overrides only, but I can add Outputs as well.

server_plugin/test/completion_contributor_test.dart
mk13: Did you mean assert NOT suggest `(foo)`?

mk13:
If we're going with the "considered available if adding decorator causes match but not available if match occurs during removal/changing of decorator" policy, then we should suggest both `[bar]` and `foo` even if technically `foo` should be there first since adding either in any order will cause it to eventually match the Directive. 

Once we get the compound function-like autocompletion we discussed in person, then we can change this logic to suggest everything at once :)

analyzer_plugin/lib/src/selector.dart
mk13: Had to change this to a non-getter since I have to pass in a parameter now.

analyzer_plugin/lib/src/selector.dart
```diff
@@ -38,28 +38,14 @@ class AndSelector extends Selector implements CompoundSelector {
...
-      } else if (selector is CompoundSelector) {
-        attributeSelectors
-            .addAll((selector as CompoundSelector).getAttributeSelectors);
-      }
-    }
-    return attributeSelectors;
-  }
+  List<AttributeSelector> getAttributeSelectors(ElementView element) =>
+      selectors
+          .expand((selector) => selector.getAttributeSelectors(element))
```
MichaelRFairhurst: ooh neat!

PR #364, opened 2017-06-06
Add exports and @deferred to readme, make an editors section, other updates

MichaelRFairhurst:
Pro tip: after reviewing diff, go to https://github.com/MichaelRFairhurst/angular2-dart-analyzer/blob/2e4881082e6675ef4233686079ba294cd4f6d04c/README.md to see how it looks

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,9 +44,13 @@ You can now use this in projects on your local system which a correctly configur
 
 ## Chart of Current Features
 
-All regular dart errors (that is to say, errors defined purely by the dart language spec) are not shown in this list.
+We plug into many editors with varying degrees of support. Anything that supports dart analysis in theory supports our plugin, but in practice that's not always the case. We do sometimes contribute back to editor plugins to make our plugin work better, so let us know what you use and what does or doesn't work.
 
-Autocomplete is available in a branch, and only in editors that support it, and maybe with bugs (many don't replace the correct content during autocomplete within html, for instance).
+Bootstrapping | Validation | Auto-Complete | Navigation | Refactoring
+--------------|------------|---------------|------------|-------------
+IntelliJ | :white_check_mark: | :white_check_mark: | :warning: some support in EAP | :no_pedestrians:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -44,42 +44,48 @@ You can now use this in projects on your local system which a correctly configur
 
 ## Chart of Current Features
 
-All regular dart errors (that is to say, errors defined purely by the dart language spec) are not shown in this list.
+We plug into many editors with varying degrees of support. Anything that supports dart analysis in theory supports our plugin, but in practice that's not always the case. We do sometimes contribute back to editor plugins to make our plugin work better, so let us know what you use and what does or doesn't work.
```
mk13: Capitalize 'dart'.

`In theory anything that supports Dart analysis also supports our plugin, but in practice that's not always the case.` 

Might sound better than plugging in 'in theory' in middle of sentence.

README.md
```diff
@@ -44,42 +44,48 @@ You can now use this in projects on your local system which a correctly configur
...
 
-Autocomplete is available in a branch, and only in editors that support it, and maybe with bugs (many don't replace the correct content during autocomplete within html, for instance).
+Bootstrapping | Validation | Auto-Complete | Navigation | Refactoring
+--------------|------------|---------------|------------|-------------
```
mk13: Do we want to add 'Quickfix' column? or will this fall under Refactoring?

MichaelRFairhurst: We could add it. I didn't know there was a difference, when I first made this. We can use :last_quarter_moon: as a "half-way" point, for when things have quickfix but not refactoring, or add another column. Partly whatever makes our OKRs best, partly whatever helps our users troubleshoot when it doesn't do what they expect...

README.md
MichaelRFairhurst: Agreed....this inspired me to revisit this paragraph, you should take another look and tell me what you think. Did I make it clearer/more to the point?

mk13: LGTM

README.md
mk13: If that's the case, it might be good to leave out 'Quickfix' until we have a better story for it and the infrastructure is at least implemented.

PR #369, opened 2017-06-08
Break circular dependency between server_plugin and analyzer_plugin

MichaelRFairhurst:
Move starter to server_plugin now that it imports server_plugin, so
that its server_plugin importing itself (which is ok) and
analyzer_plugin (which is ok), instead of it being analyzer_plugin
importing itself (which is ok) and server_plugin (which is not ok).

Circular depnedencies are OK for local development, but not publishable
(I don't think?) and not OK with bazel, which matters for shipping this
to googlers internally.

Also clean up bin/server.dart while I'm in there.


server_plugin/bin/server.dart
```diff
@@ -4,19 +4,13 @@
...
 
-import 'package:angular_analyzer_plugin/starter.dart' as ng;
-//import 'package:angular_analyzer_server_plugin/plugin.dart';
-import 'package:plugin/plugin.dart';
+import 'package:angular_analyzer_server_plugin/starter.dart' as ng;
 
-/**
- * Create and run an analysis server with Angular plugins.
- */
+///
```
liuming0: You can remove the top and bottom "///"

PR #371, opened 2017-06-09
Generator contributor autocomplete api

mk13:
Implement new Generator-Contributor plugin paradigm. Also implemented backwards compatibility for 'older' method (analysis_server).

Some minor lint-error cleanups and plugin fix as well.

Note that g3 release won't have autocomplete available for .dart 'template', but will work for .html files.


new_plugin/lib/plugin.dart
```diff
@@ -122,24 +128,15 @@ class AngularAnalysisPlugin extends ServerPlugin {
...
+  Future<ResolveResult> getResolveResultForCompletion(
+          AngularDriver driver, String path) async =>
+      new CompletionResolveResult(path);
+
+  @override
+  List<CompletionContributor> getCompletionContributors(AngularDriver driver) =>
+      driver.completionContributors;
+
+  void _loadCompletionContributors(AngularDriver driver) {
+    driver.completionContributors.add(new AngularCompletionContributor(driver));
```
bwilkerson: Needing to pass the driver in to the `AngularCompletionContributor` causes the driver to need know about contributors and causes you to have multiple instances of the contributor class. An alternative to consider is to pass the driver in as part of the `ResolveResult` that is in the `CompletionRequest` object. (Longer term I'd like `ResolveResult` to have a getter that would return the `AnalysisSession` that produced the result.)

server_plugin/lib/src/completion.dart
```diff
@@ -315,95 +312,105 @@ class TemplateCompleter {
...
 
       dartRequest.libraryElement = template.view.classElement.library;
       final memberContributor = new TypeMemberContributor();
       final inheritedContributor = new InheritedReferenceContributor();
-
-      suggestions
-        ..addAll(
-          inheritedContributor.computeSuggestionsForClass(
+      inheritedContributor
+          .computeSuggestionsForClass(
```
bwilkerson: This method is only defined on the contributor implemented in analysis_server, which implies that you still have a dependency on analysis_server.

Also, I was expecting that these contributors would be added to the list of contributors returned from `getCompletionContributors` rather than being invoked indirectly. Is there a reason why you're doing it this way?

server_plugin/lib/src/completion.dart
```diff
@@ -315,95 +312,105 @@ class TemplateCompleter {
...
         }
       }
-    } else if (target is ElementInfo) {
+      return null;
+    }
+    final replacementRangeCalculator = new ReplacementRangeCalculator(request);
+    target.accept(replacementRangeCalculator);
+    collector
+      ..offset = replacementRangeCalculator.offset
+      ..length = replacementRangeCalculator.length;
```
bwilkerson: Unless I'm missing something, the `offset` and `length` are already being set on line 315, so this should throw an exception.

mk13: Ah this has been handled. The 315 offset/length replacing happens mutually exclusive to this offset replace - so there should never be a violation.

new_plugin/lib/plugin.dart
mk13: This means we would have to change the overloadable function `getCompletionContributors(covariant Driver driver)` right? 

If I'm not mistaken - currently the only way is to store static contributors within the driver.

server_plugin/lib/src/completion.dart
mk13: This is just the first step in integrating the new architecture into our existing code. 

Ideally I would want to move these two Dart-based contributors into the contributors list generated from `getCompletionContributors` as you suggest. This would mean that our `AngularAnalysisPlugin` would have the dependency to `analysis_server` - is this fine?

MichaelRFairhurst:
@bwilkerson 

> but I think you should move all of the code

Yes! That's our next plan. We want to consolidate down to one single publishable package for external beta-ing.

server_plugin/lib/src/completion.dart
```diff
@@ -166,8 +166,12 @@ class LocalVariablesExtractor extends AngularAstVisitor {
 
 class ReplacementRangeCalculator extends AngularAstVisitor {
   CompletionRequestImpl request;
```
MichaelRFairhurst: looks like we don't need to keep this field any more

server_plugin/lib/src/completion.dart
```diff
@@ -249,21 +248,24 @@ class AngularCompletionContributor extends CompletionContributor {
...
-    final filePath = request.source.toString();
+  Future<Null> computeSuggestions(
+      CompletionRequest request, CompletionCollector collector) async {
+    final path = request.result.path;
+    final templates = await driver.getTemplatesForFile(path);
+    if (templates.isEmpty) {
+      // Todo(Max): Once fix added in plugin architecture, remove offset/length.
+      collector
+        ..offset = request.offset
+        ..length = 0;
```
MichaelRFairhurst: Hmm, do we have to do this?

server_plugin/lib/src/completion.dart
MichaelRFairhurst: We wouldn't be able to simply include them in the list, because

- we don't always run them (often the completion target is not dart)
- they require a specially set up EmbeddedDartCompletionRequest, not a regular CompletionRequest unless we're careful how its set up

I was thinking that we would wrap these contributors in ones that can translate the angular landscape into the micro dart landscape, and knows when not to run them, but otherwise just delegates to the existing behavior.

server_plugin/lib/src/completion.dart
MichaelRFairhurst: 315 runs when dart is targeted, this runs when html is targeted. Notice they are two different APIs, `ReplacementRange.compute(request)` vs `new ReplacementRangeCalculator(request).accept(target)`

server_plugin/lib/src/completion.dart
```diff
@@ -315,95 +316,105 @@ class TemplateCompleter {
...
         if (target.parent != null || target.parent is! DocumentInfo) {
-          suggestTransclusions(target.parent, suggestions);
+          suggestTransclusions(target.parent, collector);
         }
       }
     } else if (target is AttributeInfo && target.parent is TemplateAttribute) {
       // `let foo`. Nothing to suggest.
       if (target is TextAttribute && target.name.startsWith("let-")) {
-        return suggestions;
+        return null;
```
MichaelRFairhurst: I think this should just be `return;`

server_plugin/lib/src/embedded_dart_completion_request.dart
```diff
@@ -1,25 +1,26 @@
...
 import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/type.dart';
 import 'package:analyzer/file_system/file_system.dart';
-import 'package:analyzer_plugin/src/utilities/completion/completion_target.dart';
-import 'package:analyzer_plugin/src/utilities/completion/optype.dart';
-
 import 'package:analyzer/dart/ast/ast.dart';
 
-class EmbeddedDartCompletionRequest implements DartCompletionRequest {
+class EmbeddedDartCompletionRequest implements old_core.DartCompletionRequest {
```
MichaelRFairhurst: Yeah, ideally we wouldn't need old_core any longer :/

server_plugin/lib/starter.dart
```diff
@@ -109,22 +112,27 @@ class Starter {
...
-              null, // AnalysisResult - unneeded for AngularCompletion
-              server.resourceProvider,
-              source,
-              params.offset,
-              performance);
+            null, // AnalysisResult - unneeded for AngularCompletion
+            server.resourceProvider,
+            source,
+            params.offset,
+            performance,
```
MichaelRFairhurst: trailing comma

server_plugin/lib/starter.dart
```diff
@@ -109,22 +112,27 @@ class Starter {
...
+          await completionContributor.computeSuggestions(newRequest, collector);
+          final suggestions = collector.suggestions;
+
           completionHandler
             ..sendCompletionNotification(
-                completionId,
-                completionRequest.replacementOffset,
-                completionRequest.replacementLength,
-                suggestions)
+                completionId, collector.offset, collector.length, suggestions)
```
MichaelRFairhurst: woo backwards compatibility!

new_plugin/lib/plugin.dart
bwilkerson: Yes, we need to figure out this `ResolveResult` mess.

But modulo that, what I was thinking is that if `getResolveResultForCompletion` returned a subclass of `ResolveResult` that also knew the `AngularDriver`, then it could be passed in (via `CompletionRequest`) to each of the contributors, making it unnecessary to pass it in when creating the `AngularCompletionContributor`.

server_plugin/lib/src/embedded_dart_completion_request.dart
bwilkerson: That's not an ideal, that's a strict requirement. No library in `new_plugin` can import anything, directly or indirectly, from `analysis_server`, because `analysis_server` isn't published and therefore won't be found at runtime.

server_plugin/lib/src/completion.dart
bwilkerson: > Ideally I would want to move these two Dart-based contributors into the contributors list generated from `getCompletionContributors` as you suggest. This would mean that our `AngularAnalysisPlugin` would have the dependency to analysis_server - is this fine?

No. No library in `new_plugin` can import, directly or indirectly, any library from `analysis_server`. I created versions of these two contributors in `analyzer_plugin/lib/utilities/completion` that you can use in place of the ones in `analysis_server`.

mk13: Sounds good, I'll transition the import settings to use the `analyzer_plugin` version instead.

mk13:
Code is temporarily in a broken (and messy) state - primarily for visibility in still on-going changes.

server_plugin/lib/src/completion.dart
```diff
@@ -231,11 +229,217 @@ class ReplacementRangeCalculator extends AngularAstVisitor {
...
+        final dartResolveResult = new NgResolveResult(request.result.path, [],
+            libraryElement: libraryElement, typeProvider: typeProvider);
+        final dartRequest = new CompletionRequestImpl(
+            request.resourceProvider, dartResolveResult, request.offset);
+
+        await super
+            .computeSuggestions(dartRequest, collector, entryPoint: entryPoint);
+
+        if (collector.suggestionsLength != initialSuggestionLength &&
+            !collector.offsetIsSet) {
```
bwilkerson: Doesn't line 276 guarantee that the offset will be set by the time we get here?

server_plugin/lib/src/completion.dart
```diff
@@ -231,11 +229,217 @@ class ReplacementRangeCalculator extends AngularAstVisitor {
...
+          }
+        }
+
+        if (collector.suggestionsLength != initialSuggestionLength &&
+            !collector.offsetIsSet) {
+          final range =
+              new ReplacementRange.compute(request.offset, completionTarget);
+          collector
+            ..offset = range.offset
+            ..length = range.length;
```
bwilkerson: Given that all of the contributors will always be run, do we really need to duplicate this logic in each one?

server_plugin/lib/src/completion.dart
mk13: Ah yes :) The code clean-up is still in the works haha.

server_plugin/lib/src/completion.dart
mk13: Theoretically - yes. These are added as barriers to ensure that 1) 'override' error of offset and length is never triggered and 2) offset and length is at least set only once when needed (suggestions.isNotEmpty). 

I didn't feel comfortable leaving these contributors in a potentially dangerous state where if the contributors are re-arranged or removed (for whatever reason), the code won't break.

new_plugin/lib/plugin.dart
mk13: sgtm

analyzer_plugin/lib/src/angular_driver.dart
```diff
@@ -50,6 +51,7 @@ class AngularDriver
   FileTracker _fileTracker;
   final lastSignatures = <String, String>{};
   bool _hasAngularImported = false;
+  final completionContributors = <CompletionContributor>[];
```
mk13: Acked.

new_plugin/lib/plugin.dart
```diff
@@ -122,24 +128,15 @@ class AngularAnalysisPlugin extends ServerPlugin {
...
-    final analysisResult = await driver.dartDriver.getResult(filePath);
-    final contributor = new AngularCompletionContributor(driver);
-    final performance = new CompletionPerformance();
-    final fileSource = resourceProvider.getFile(filePath).createSource();
-    final request = new CompletionRequestImpl(analysisResult, resourceProvider,
-        fileSource, parameters.offset, performance);
-    final suggestions = await contributor.computeSuggestions(request);
-    return new plugin.CompletionGetSuggestionsResult(
-        request.replacementOffset, request.replacementLength, suggestions);
+  Future<ResolveResult> getResolveResultForCompletion(
```
mk13: I'm not 100% aware of how covariant works, but doesn't this only apply to the parameter driver and not the return type?

server_plugin/lib/starter.dart
mk13: Intentional for nice dart-formatting!

mk13:
Ping @MichaelRFairhurst for another review before merging.

server_plugin/lib/src/completion.dart
MichaelRFairhurst: I think it'd be OK if it was split out into its own contributor. Alternatively, we could use a mixin, or inheritance if these are changed to no longer extend the base dart contributors

MichaelRFairhurst: Another thought, if these contributors _use_ the dart contributors, rather than _extending_ them, then we can actually have this be one single contributor class so that it can do this logic just once and not be problematic to move it around in the contributors list.

We could even, then, make our own little contributors framework inside of the contributor; this can be a contributor that finds a dartSnippet and sets the completion length; then it can have an array of special contributors which run off of the dartSnippet, and it can use each of them to get suggestions.

```
class EmbeddedDartCompletionContributor implements CompletionContributor {
  final List<DartSnippetBasedCompletionContributor> contributors;
  computeSuggestions(...) {
    findDartSnippet();
    setCompletionOffsetAndLength();
    for (final contributor in contributors) {
     await contributor.computeSuggestions(dartSnippet);
    }
    moreStuffAfterAllHaveRun();
  }
}
class DartSnippetBasedCompletionContributor {
  computSuggestions(AstNode dartSnippet);
}

class NgTypedMemberContributor extends DartSnippetBasedCompletionContributor {
  ...
}
class NgInheritedMemberContributor extends DartSnippetBasedCompletionContributor {
  ...
}
...

class Plugin {
  get contributors => [new EmbeddedDartContributor([new NgTypedMemberContributor(), new NgInheritedMemberContributor()])];
  ...
}
```

now if you you try to run `NgTypedMemberContributor` where a regular `Contributor` is expected, the code has type errors telling you you messed up

server_plugin/lib/src/completion.dart
```diff
@@ -487,6 +630,13 @@ class TemplateCompleter {
       suggestHtmlTags(template, collector);
       suggestTransclusions(target.parent, collector);
     }
+
+    if (collector.suggestionsLength != initialSuggestionsCount &&
```
MichaelRFairhurst: Hmm, maybe one contributor should handle setting the offset/length, which can tell whether to use the dart strategy or the html strategy. Then we won't need to track suggestionsLength at all

server_plugin/lib/starter.dart
```diff
@@ -109,8 +110,11 @@ class Starter {
             driverPath) {
           final driver = angularDrivers[driverPath];
 
-          final completionContributor =
-              new AngularCompletionContributor(driver);
+          final contributors = <CompletionContributor>[
+            new AngularCompletionContributor(driver),
```
MichaelRFairhurst: did we end up using the driver?

new_plugin/lib/plugin.dart
mk13: > Debatably, this make the typesystem unsound. However, "soundness" isn't strictly defined...the argument the dart team makes is that, as long as an error is raised predictably when this happens, the typesystem is still arguably sound.

Good enough argument for me.

mk13: Unfortunately we will still fall into the need to cast it anyways. 

If you look at `completion.dart` under `AngularCompletionContributor`'s `computeSuggestions`, we cast it there. However, we grab this object directly from `CompletionRequest` which is loaded with the `ResolveResult` through `getResolveResultForCompletion`. We never call `getResolveResultForCompletion` directly - but we do call request.result. `getResolveResultForCompletion` is called directly from `handleCompletionGetSuggestions` which we aren't overloading. 

As much as it's a great idea - it wouldn't have any pay off and just leads our code into a potentially 'unsound' state since we don't fully control the return value of this function call; we just piggy back off of `CompletionMixin`'s `handleCompletionGetSuggestion` that we aren't overloading.

server_plugin/lib/src/completion.dart
```diff
@@ -230,13 +227,211 @@ class ReplacementRangeCalculator extends AngularAstVisitor {
...
 
+/// Extension of [TypeMemberContributor] to allow for Dart-based
+/// completion within Angular context. Triggered in [StatementsBoundAttribute],
+/// [ExpressionsBoundAttribute], [Mustache], and [TemplateAttribute]
+/// on member variable completion.
+class NgTypeMemberContributor extends TypeMemberContributor {
+  @override
+  Future<Null> computeSuggestions(
+      CompletionRequest request, CompletionCollector collector,
+      {AstNode entryPoint: null}) async {
```
mk13: Nope - I knew I forgot something in the code cleanup.

server_plugin/lib/src/completion.dart
```diff
@@ -230,13 +227,211 @@ class ReplacementRangeCalculator extends AngularAstVisitor {
...
+          template.view.templateOffset <= request.offset &&
+              request.offset < template.view.end;
+      if (!isFromHtmlFile && !isFromValidDartTemplate) {
+        continue;
+      }
+      final initialSuggestionLength = collector.suggestionsLength;
+      final typeProvider = template.view.component.classElement.enclosingElement
+          .enclosingElement.context.typeProvider;
+      final target = findTarget(request.offset, template.ast);
+      final extractor = new DartSnippetExtractor()..offset = request.offset;
```
mk13: If this is to eliminate the redundancy of calculating both of these variables, I would much rather instead condense the two inherited contributors into a single contributor that keeps the two as a property (as mentioned in your 'composition over inheritance') post. 

I'm not much of having a bidirectional connection between contributor and result object. Ideally any request and result, once formed, shouldn't be mutated in any way shape or form by the contributors.

server_plugin/lib/src/completion.dart
```diff
@@ -230,13 +227,211 @@ class ReplacementRangeCalculator extends AngularAstVisitor {
...
+
+    for (final template in templates) {
+      // Check if this template is valid.
+      final isFromHtmlFile = template.view.templateUriSource != null;
+      final isFromValidDartTemplate =
+          template.view.templateOffset <= request.offset &&
+              request.offset < template.view.end;
+      if (!isFromHtmlFile && !isFromValidDartTemplate) {
+        continue;
+      }
```
mk13: That's right - I've moved the templates into the result object. I'll perform the filter there.

server_plugin/lib/starter.dart
mk13: Yes to calculate StandardHtml content - but we can potentially move this to the resolve object.

server_plugin/lib/src/completion.dart
mk13: Sure I can try that.

server_plugin/lib/src/completion.dart
mk13: I'm in favor of the solution posted below about condensing them.

mk13:
Discussed in person - current system doesn't allow for a perfect single solution. Points that need to be resolved: 
* A need for some sort of "pre-processing" step that is guaranteed to occur first without contributing anything.
* A mechanism to allow for shared information to flow through all contributors without having to unnecessarily overload the Result object. Ideally the Result object should be more 'static' while the said container object for shared information is mutable and short-lived.

One possible suggestion is to remove `ResolveResult` from piggy backing off of request. Instead, we create a new class along the names of `Context` or `Container`. This `Container` is extendable by plugin implementers. 

On the Dart analysis side, this container should be extended as something along the lines of `DartContainer` and import the `ResolveResult`. In addition, change `ResolveResult` to be specific to this type only and not to the generic `Container` object. 

This way should users want to build off of the groundwork done by Dart - they can pick up the contributor logic from here.

server_plugin/lib/src/completion.dart
mk13: This will have to be implemented once we re-design the ResolveResult and potential container architecture as listed below.

Issue #368, opened 2017-06-07
<template ngFor let-item [ngForOf]="data"> results in an error


zoechi:
```
    <template ngFor let-item [ngForOf]="dataSet.bundle.entry">
      <template [ngTemplateOutlet]="rowTemplate" [ngTemplateOutletContext]="item.resource"></template>
                                                                            ^^^^
    </template>
```

```
error: Undefined name 'item'. (undefined_identifier at [fhir_webui] lib/src/webui_generated/data_set_overview/data_set_overview_component.html:12)
```

MichaelRFairhurst:
@maxkim do you have any ideas on why this is happening? Something about
nesting that's not working?

mk13:
It might be something within the recursive scoper - will take a look.

mk13:
Resolved - please raise another issue if a related issue arises.

Issue #367, opened 2017-06-07
foo in <template #foo> returns a TemplateRef not a TemplateElement


zoechi:
```
<template #propertyListItemTemplate let-listItem="item">
```

with 

```
    <list-component 
           [template]="propertyListItemTemplate"
```

and

```dart
class ListComponent {
  @Input()
  TemplateRef template;
```

causes 
```
warning: Attribute value expression (of type TemplateElement) is not assignable to component input (of type TemplateRef) (input_binding_type_error at [fhir_designer] lib/webui/component_property_editor/component_property_editor_component.html:15)
```

but I think this should not cause a warning and the code is working as intened.

MichaelRFairhurst:
Yes, I do believe you're right @zoechi.

@maxkim can you take a look? We have the TemplateRef ClassElement available
via standardAngular.templateRef

zoechi:
In TS land they split the functionality. They have now `<ng-template>` for above use case and `<template>` is a plain old `TemplateElement`.
Perhaps it's time to get that into Dart Angular as well.
It limits for example using Polymer with Angular.

zoechi:
@matanlurey any plans for `<ng-template>` in Dart Angular?

mk13:
Taking a look.

mk13:
Resolved - please raise another issue if a related issue arises.

PR #372, opened 2017-06-10
Issue 368

mk13:
Fix for issue #367 & #368 

`test_hashRef_templateElement()` should have picked the issue for 367 - need to figure out what went wrong in test case configuration.

EDIT: Fixed test case issue.

### DIFF
--- a/analyzer_plugin/lib/src/resolver.dart
+++ b/analyzer_plugin/lib/src/resolver.dart
@@ -163,7 +163,8 @@ class TemplateResolver {
             templateSource,
             typeProvider,
             dartVarManager,
-            errorListener))
+            errorListener,
+            standardAngular))
         // Load $event into the scopes
         ..accept(new PrepareEventScopeVisitor(
             standardHtmlEvents,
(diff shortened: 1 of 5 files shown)


analyzer_plugin/lib/src/resolver.dart
```diff
@@ -488,7 +492,10 @@ class PrepareScopeVisitor extends AngularScopeVisitor {
           // Find the corresponding Component to assign reference to.
           for (final directive in directives) {
             if (directive is Component) {
-              final classElement = directive.classElement;
+              var classElement = directive.classElement;
```
MichaelRFairhurst: Hmm, I'm wondering if we should have a `TemplateElement` `Component` at all. Of course, we'd still have to make this check here, just, using the element name instead.

Maybe the `TemplateElement` component could have the `TemplateRef` class for its class element? Not sure if that strategy would be laying a trap for later on.

mk13: We would probably have to touch base with the Angular2 crew to figure out the best strategy here. 

Unfortunately for us we're kind of at the mercy of dartium_html on how they decide to define the `<template>` element. Zoechi briefly touched on this within the issue #367 on this issue. 

I don't much like the idea of doing a hard string comparison on 'TemplateElement' - but not sure what else we can do at this point. We don't exactly import the angular2 library directly without our plugin so it's hard to get a hard class type; instead we're at the mercy of Strings for the moment. 

Pinging @matanlurey - How do you think we should approach this? Is there any planned changed for better synergy between `TemplateElement` and `TemplateRef`?

PR #374, opened 2017-06-13
Handle 'exports' feature -- restrict scope to exported scope & class.

MichaelRFairhurst:
Looks like previously all accessors, top levels, classes, enums, and
more were available in templates already.

Therefore the feature is more about collecting exports (and validating
how they are defined), and then restricting the available top levels to
what was available.

Looks like the way to restrict the imports requires creating a layer
between the resolver visitors and LibraryElements,
CompilationUnitElements, and ImportElements and Namespaces. Created a
new folder 'facade' for these wrappers. Unfortunately its a lot of
fragile boilerplate, which may doom that part of this implementation to
maintenance and versioning woes in the future. A possible solution to
this is to create an easier-to-maintain interface which resolver can
accept which we can then fulfill based on the exports array, while the
analyzer can fulfill it based on the Element model.

Alternatively, this could be done with mirrors, except that's
deprecated and all that.

Element model solution has some quirks:
- CompilationUnitElements of a LibraryElement's (the definingUnit)
determines the identifiers which don't require importing.
- LibraryElement's imports which carry prefixes
- LibraryElmeent's Namespaces for imported units

A few things are not carefully prevented due to difficulty of
implementation:
- getAncestor may skip over the facades
- elements which are exported may be returned as part of a namespace or
accessor, etc, of a unit or library to which it doesn't actually belong
- going into children may skip over the facades
- enclosingElement may skip over the facades
- LibraryElement.getType(String) may return unexported names
- likely much more because the API is very very very wide

However, as the tests pass, these issue may not be real issues. Likely,
as long as the issues result in suppressing errors rather than false
errors, the priority of any such bugs will be low.


analyzer_plugin/lib/src/facade/exports_compilation_unit_element.dart
```diff
@@ -0,0 +1,183 @@
...
+
+  @override
+  String computeDocumentationComment() => _wrappedUnit
+      .computeDocumentationComment(); // ignore: deprecated_member_use
+
+  @override
+  CompilationUnit computeNode() => _wrappedUnit.computeNode();
+
+  @override
+  Element/*=E*/ getAncestor/*<E extends Element >*/(
```
hterkelsen: are you able to use actual generics here or do you still need to support old versions of Dart?

analyzer_plugin/lib/src/facade/exports_compilation_unit_element.dart
```diff
@@ -0,0 +1,183 @@
...
+  @override
+  /*=T*/ accept/*<T>*/(ElementVisitor<dynamic/*=T*/ > visitor) =>
+      _wrappedUnit.accept(visitor);
+
+  @override
+  bool get hasLoadLibraryFunction => _wrappedUnit.hasLoadLibraryFunction;
+
+  @override
+  List<PropertyAccessorElement> get accessors =>
+      new List<PropertyAccessorElement>.from((_component?.view?.exports ?? [])
```
hterkelsen: maybe make `_component?.view?.exports ?? []` a field to avoid repeating it

analyzer_plugin/lib/src/facade/exports_compilation_unit_element.dart
```diff
@@ -0,0 +1,183 @@
...
+          .where(_fromThisUnit)
+          .map((export) => export.element)
+          .where((element) => element is PropertyAccessorElement));
+
+  @override
+  List<ClassElement> get enums =>
+      new List<ClassElement>.from((_component?.view?.exports ?? [])
+          .where(_fromThisUnit)
+          .map((export) => export.element)
+          .where((element) => element is ClassElement));
```
hterkelsen: shouldn't this check that the element is an enum?

analyzer_plugin/lib/tasks.dart
```diff
@@ -366,6 +368,14 @@ class AngularWarningCode extends ErrorCode {
           ' not the first match. Use (Content or View)Children to allow'
           ' multiple matches.');
 
+  /// An error code indicating that the exports array got a non-identifier
+  static const EXPORTS_MUST_BE_PLAIN_IDENTIFIERS = const AngularWarningCode(
+      'EXPORTS_MUST_BE_PLAIN_IDENTIFIERS', 'Exports must be plain identifiers');
+
+  /// An error code indicating that the exports array got a non-identifier
```
hterkelsen: this comment is copied from EXPORTS_MUST_BE_PLAIN_IDENTIFIERS

analyzer_plugin/lib/src/view_extraction.dart
```diff
@@ -185,6 +188,50 @@ class ViewExtractor extends AnnotationProcessorMixin {
...
+          // Check dupes BEFORE separating the identifier & prefix
+          if (alreadyExported.contains(name)) {
+            errorReporter.reportErrorForNode(
+                AngularWarningCode.DUPLICATE_EXPORT, item, [name]);
+          }
+          alreadyExported.add(name);
+
+          // Separate the identifier & prefix
+          var prefix = '';
+          if (item is ast.PrefixedIdentifier) {
```
hterkelsen: not all prefixed identifiers are allowed. you should report an error if the Prefix is not a library prefix (ie, `MyClass.staticField` is not allowed, you must export `MyClass` instead). the angular compiler will report an error in this case

bwilkerson:
I don't think I have the context to review this. If there are specific areas you'd like me to look at I'm happy to do so.

analyzer_plugin/lib/src/facade/exports_compilation_unit_element.dart
MichaelRFairhurst: The signature is copied from the class I'm proxying...I'm actually not sure what the consequences are of turning this into a real generic given that the parent class still uses the comment syntax. No, it doesn't need to support older versions of dart

analyzer_plugin/lib/src/facade/exports_compilation_unit_element.dart
MichaelRFairhurst: Will do

analyzer_plugin/lib/src/facade/exports_compilation_unit_element.dart
MichaelRFairhurst: Good catch!

analyzer_plugin/lib/tasks.dart
MichaelRFairhurst: Will fix

analyzer_plugin/lib/src/view_extraction.dart
MichaelRFairhurst: I'll add a test case -- this should be handled, with an emphasis on _should be_, therefore, needs tests :)

analyzer_plugin/lib/src/facade/exports_compilation_unit_element.dart
MichaelRFairhurst: Done

analyzer_plugin/lib/src/facade/exports_compilation_unit_element.dart
MichaelRFairhurst: Done

analyzer_plugin/lib/tasks.dart
MichaelRFairhurst: Done

analyzer_plugin/lib/src/view_extraction.dart
MichaelRFairhurst: Thanks for this catch. I could have sworn prefixed identifier was magically only for true prefixes. Also based my approach off of the angular pattern as best I could, but we have to be careful not to rely on resolved ASTs so I found a way to perform your same check by using summaries only, after this check, when the directive is linked later.

analyzer_plugin/lib/src/directive_linking.dart
```diff
@@ -208,22 +209,46 @@ class ExportLinker {
...
+        continue;
+      }
+
       export.element = _scope.lookup(getIdentifier(export), null);
     }
   }
 
+  /// Only report false for known non-import-prefix prefixes, the rest get
+  /// flagged by the dart analyzer already.
+  bool hasWrongTypeOfPrefix(ExportedIdentifier export) {
```
hterkelsen: can't you just do `prefixElement is! PrefixElement`? That's what we do in the angular compiler.

MichaelRFairhurst:
Switching Brian out for @scheglov since Brian's going on vacation.

@scheglov, to answer Brian's question, the review I need here is related to the 'facade' directory, which I'm using to reduce the library/compilation unit's namespace down to what has been specified using `exports: [constTerms, goHere, sometimes.withPrefixes]`.

Its fragile since it relies on wrapping the Element model, which is a large API that's difficult to guarantee I've correctly wrapped in full. I'd be happy to change it to a different solution, including talking about analyzer changes that might require.

analyzer_plugin/lib/src/directive_linking.dart
MichaelRFairhurst: If I return false for when prefixElement is null, then I'll report an error for mispelled prefixes on top of the error that dart analysis already produces. I'll add a test for this to confirm

MichaelRFairhurst: test case added; indeed, if I remove the null check then our code triggers secondary errors where they are already reported.

Funnily, though, a misspelled prefix already gives off three errors within the analyzer itself :man_shrugging: not sure if that makes this extra important (four would be pretty crazy) or extra low priority given that its not clean no matter how I do it! :)

analyzer_plugin/test/angular_driver_test.dart
```diff
@@ -3379,6 +3498,133 @@ class ComponentA {
...
   }
 
+  // ignore: non_constant_identifier_names
+  Future test_hasExports() async {
+    final code = r'''
+import 'package:angular2/angular2.dart';
+
+const String foo = 'foo';
+const int bar = 2;
+
```
mk13: It might be nice to add a test case for function returns that that the function returntype matches as expected - similar to testing if the variable element types match.

MichaelRFairhurst: Ah, good idea, will do

server_plugin/lib/src/completion.dart
```diff
@@ -978,18 +1166,6 @@ class TemplateCompleter {
     });
   }
 
-  void addLocalVariables(CompletionCollector collector,
```
mk13: This was removed but we're still using it at line 355?

server_plugin/test/completion_contributor_test.dart
```diff
@@ -3294,4 +3294,195 @@ class CustomTemplateDirective {
...
+
+    addTestSource('{{^}}');
+
+    await resolveSingleTemplate(dartSource);
+    await computeSuggestions();
+    expect(replacementOffset, completionOffset);
+    expect(replacementLength, 0);
+    assertSuggestGetter('myAccessor', 'int');
+    assertSuggestFunction('myFunction', 'int');
+    assertSuggestClass('MyClass');
```
mk13: Should this also be suggesting `MyComp`?

MichaelRFairhurst: Yeah, I put it in the next test...I didn't feel like it made sense to assert that it suggested the current component in `test_completesExports` when the current component isn't an export.

mk13: Ah gotcha. Fair enough.

analyzer_plugin/test/angular_driver_test.dart
MichaelRFairhurst: Fixed!

Also realized that I wrote these tests before I knew that classes always exported themselves. So I actually changed that (a component exporting itself) to an error with tests, and changed these tests to no longer do that.

Feel free to review that change too, though I think its safe & small

PR #375, opened 2017-06-15
Ref value autocomplete

mk13:


server_plugin/lib/src/completion.dart
```diff
@@ -890,6 +898,28 @@ class TemplateCompleter {
     }
   }
 
+  void suggestRefValues(
+      List<DirectiveBinding> directives, CompletionCollector collector) {
+    // Keep map of all 'exportAs' name seen. Don't suggest same name twice.
+    // If two directives share same exportAs, still suggest one of them
+    // and if they use this, and error will flag - let user resolve
+    // rather than not suggesting at all.
+    final seen = new HashSet<String>();
```
liuming0: You can use Set here.

server_plugin/lib/src/completion.dart
```diff
@@ -890,6 +898,28 @@ class TemplateCompleter {
     }
   }
 
+  void suggestRefValues(
+      List<DirectiveBinding> directives, CompletionCollector collector) {
+    // Keep map of all 'exportAs' name seen. Don't suggest same name twice.
```
MichaelRFairhurst: Just curious -- are there any common examples where this would be suggested twice?

server_plugin/lib/src/completion.dart
```diff
@@ -890,6 +898,28 @@ class TemplateCompleter {
...
+    // If two directives share same exportAs, still suggest one of them
+    // and if they use this, and error will flag - let user resolve
+    // rather than not suggesting at all.
+    final seen = new HashSet<String>();
+    for (final directive in directives) {
+      final exportAs = directive.boundDirective.exportAs;
+      if (exportAs != null && exportAs.name.isNotEmpty) {
+        final exportAsName = exportAs.name;
+        if (exportAsName.isNotEmpty && !seen.contains(exportAsName)) {
```
MichaelRFairhurst: already checked exportAs.name.isNotEmpty

MichaelRFairhurst:
Oh! Can you also update the README to go with this?

server_plugin/lib/src/completion.dart
mk13: Probably not - but currently there is not indicated from the Angular compiler when this occurs; it just silently fails. It would be nice if our plugin can at least provide a warning in such cases rather than leaving Angular developers scratching their heads in the rare cases where this does happen.

server_plugin/lib/src/completion.dart
mk13: Good catch! Thanks!

mk13:
Ping @MichaelRFairhurst if the README looks good to you.

README.md
```diff
@@ -73,6 +73,7 @@ Template syntax | Validation | Auto-Complete | Navigation | Refactoring
 `<my-cmp></my-cmp>` | :white_check_mark: existence of directive |:white_check_mark: | :x: | :x:
 `<my-cmp [(title)]="name">` | :white_check_mark: soundness of expression, existence of `title` input and `titleChange` output on directive or component with proper type | :white_check_mark: | :x: | :x:
 `<video #movieplayer ...></video><button (click)="movieplayer.play()">` | :white_check_mark: type of new variable tracked and checked in other expressions | :white_check_mark: | :x: | :x:
+`<video directiveWithExportAs #moviePlayer="exportAsValue">` | :white_check_mark: existence of exportAs value checked within bound directives | :white_check_mark: | :x: | :x:
```
MichaelRFairhurst: Looks good, can you also update the `#myForm="ngForm"` line 91?

server_plugin/lib/src/completion.dart
```diff
@@ -904,18 +904,16 @@ class TemplateCompleter {
...
       if (exportAs != null && exportAs.name.isNotEmpty) {
         final exportAsName = exportAs.name;
-        if (exportAsName.isNotEmpty && !seen.contains(exportAsName)) {
-          seen.add(exportAsName);
-          collector.addSuggestion(_createRefValueSuggestion(
-              exportAs,
-              DART_RELEVANCE_DEFAULT,
-              _createRefValueElement(exportAs, protocol.ElementKind.LABEL)));
-        }
+        seen.add(exportAsName);
```
MichaelRFairhurst: Looks like you removed the dedupe check in addition to the extra isNotEmpty check

mk13: Speaking of which, I should add the dedupe test case.

PR #386, opened 2017-06-25
Request object change

mk13:
Changing Completion architecture to use new Request-based paradigm. 
TypeMemberContributor and InheritedReferenceContributor now uses composition over inheritance.
Offset & Length calculation is now done in its own contributor.


mk13:
For now I think this is as clean as we can go. And we were able to separate some of the repeated data objects that we pass through our contributors and recycle some code to reduce redundant code :)

I think for now this is a great step forward. At least in terms of Completion service, this is a fairly clean approach and I suspect the other services will be just as clean - but won't know for certain until we fully implement them. 

Fixes might potentially require some changes, but I don't think it's going to be a huge change if there is.

new_plugin/lib/src/completion_request.dart
```diff
@@ -0,0 +1,199 @@
...
+
+  @override
+  void checkAborted() {
+    if (_aborted) {
+      // ignore: only_throw_errors
+      throw new AbortCompletion();
+    }
+  }
+
+  AstNode _calculateDartSnippet(Template template) {
```
MichaelRFairhurst: this looks great...I'm loving this...

server_plugin/lib/src/completion.dart
```diff
@@ -264,58 +148,91 @@ class ReplacementRangeCalculator implements AngularAstVisitor {
...
+  @override
+  ResultState get state => null;
+
+  @override
+  Uri get uri => null;
+
+  _ResolveResultShell(this.path, {this.libraryElement, this.typeProvider});
+}
+
+class NgOffsetLengthContributor extends CompletionContributor {
```
MichaelRFairhurst: this is also super nice

PR #388, opened 2017-06-29
Expose pipe

mk13:
Exposes pipes : @Pipe declared classes, @Component(... pipes: const [...]), and linking between the two. 

Validation of pipe will require some dependency management of the shared analyzer dependency within angular_ast and this plugin. One possible long term solution is to utilize angular_ast's expression parsing and completely remove the equation from our end so we never have any overlaps and we can simply just use angular_ast to return proper expression containers - but this is a significant change since we do a lot of expression parsing within the plugin.


analyzer_plugin/lib/src/model.dart
```diff
@@ -347,6 +348,20 @@ class OutputElement extends AngularElementImpl {
       'OutputElement($name, $nameOffset, $nameLength, $getter)';
 }
 
+class Pipe {
+  final String pipeName;
+  final int pipeNameOffset;
+  final dart.ClassElement classElement;
+  final bool isPure;
+
+  dart.InterfaceType requiredArgumentType;
```
MichaelRFairhurst: I think these can be just dart.Type. For instance, a pipe might return "dynamic" which isn't an interface

analyzer_plugin/lib/src/pipe_extraction.dart
```diff
@@ -0,0 +1,124 @@
...
+        final constantEvaluation =
+            calculateStringWithOffsets(pipeNameExpression);
+        if (constantEvaluation != null && constantEvaluation.value is String) {
+          pipeName = (constantEvaluation.value as String).trim();
+          pipeNameOffset = pipeNameExpression.offset;
+        }
+      }
+      if (isPureExpression != null) {
+        final isPureValue =
+            isPureExpression.accept(new OffsettingConstantEvaluator());
```
MichaelRFairhurst: I don't think you want OffsettingConstantEvaluator, that's for taking things like `"foo" + "bar"` and turning it into ` foo     bar` so that its a single string where the offsets match.

I think there's a ConstantEvaluator class which OffsettingContantEvaluator extends which you can use.

analyzer_plugin/lib/src/pipe_extraction.dart
```diff
@@ -0,0 +1,124 @@
...
+          isPure = isPureValue;
+        }
+      }
+      if (pipeName == null) {
+        errorReporter.reportErrorForNode(
+            AngularWarningCode.PIPE_SINGLE_NAME_REQUIRED, node);
+      }
+
+      // Check if 'extends PipeTransform' exists.
+      final superType = _currentClassElement.supertype;
```
MichaelRFairhurst: Looks like some people implement PipeTransform instead of extending it. In theory you could also use a mixin, or extend a class that extends PipeTransform.

I think the best way to detect this is `_standardAngular.pipeTransform.type.isSupertypeOf(_currentClassElement.type)`

analyzer_plugin/lib/src/pipe_extraction.dart
```diff
@@ -0,0 +1,124 @@
...
+
+  /// Looks for a 'transform' function, and if found, finds all the
+  /// important type information needed for resolution of pipe.
+  Pipe _loadTransformInformation(Pipe pipe) {
+    final classElement = pipe.classElement;
+    if (classElement == null) {
+      return pipe;
+    }
+    final transformMethod = classElement.getMethod('transform');
+    if (transformMethod == null) {
```
MichaelRFairhurst: Also to check: can transform be generic? Is it abstract? Static?

analyzer_plugin/lib/src/pipe_extraction.dart
```diff
@@ -0,0 +1,124 @@
...
+    final parameters = transformMethod.parameters;
+    if (parameters == null || parameters.isEmpty) {
+      errorReporter.reportErrorForElement(
+          AngularWarningCode.PIPE_TRANSFORM_REQ_ONE_ARG, transformMethod);
+    }
+    for (final parameter in parameters) {
+      // If named or positional
+      if (parameter.parameterKind.ordinal > 0) {
+        errorReporter.reportErrorForElement(
+            AngularWarningCode.PIPE_TRANSFORM_NO_NAMED_ARGS, parameter);
```
MichaelRFairhurst: ah great edge case!

analyzer_plugin/lib/tasks.dart
```diff
@@ -134,6 +140,12 @@ class AngularWarningCode extends ErrorCode {
       'The type "{0}" is included in the directives list, but is not a'
       ' directive');
 
+  /// An error code indicating that the value of type is not a Pipe.
+  /// An error code indicating that the value of an expression is not a string.
```
MichaelRFairhurst: extra comment

angular_analyzer_plugin/lib/src/pipe_extraction.dart
```diff
@@ -73,9 +74,11 @@ class PipeExtractor extends AnnotationProcessorMixin {
       }
 
       // Check if 'extends PipeTransform' exists.
-      final superType = _currentClassElement.supertype;
-      if (superType == null ||
-          superType != _standardAngular.pipeTransform.type) {
+      var allSupertypes = _currentClassElement.allSupertypes ?? [];
+      allSupertypes = allSupertypes
+          .where((t) => _standardAngular.pipeTransform.type.isSupertypeOf(t))
```
MichaelRFairhurst: Ah interesting! Clever way of doing this. In fact, solves more problems than it has to.

Subtypes are `<:` subtypes. So if any of T's supertypes `<:` PipeTransform, and T's supertypes `<:` T, then T `<:` PipeTransform.

So you can just do, `if (!standardAngular.pipeTransform.type.isSupertypeOf(_currentClassElement.type)) { error }` I think

PR #389, opened 2017-07-11
Fix build: analyzer_plugin driver changes, no MethodElement locals

MichaelRFairhurst:
Not sure what method element local variables were doing before.
Removing that code seemed to cause no changes in behavior.

Removing the driver in place for a path was simple enough.


zoechi:
Is this somehow related to dart-lang/sdk#30135 ?

MichaelRFairhurst:
Yes! Feel free to use my fix here before its merged.

PR #390, opened 2017-07-13
Opt in to analysis in new plugin arch via yaml file

MichaelRFairhurst:
Since `ServerPlugin` expects a driver and calls methods on it behind
the scenes, return a `NoopDriver` for the analysis roots that don't
have angular enabled.

Our own code has to detect the NoopDrivers and bail out early, but
that's easy enough to add.

Expects your `analysis_options.yaml` to have:

```
angular:
  enabled: true
```

In order to run our analysis.

New dependency on package:yaml, and handle edge cases like file not
actually existing, etc.

Tests for that as well, with mocks thanks to a little FileService that
gets injected.


angular_analyzer_plugin/lib/src/file_service.dart
```diff
@@ -0,0 +1,6 @@
+import 'dart:io';
+
+/// Class for testing. This can be mocked to provide mock files.
+class FileService {
```
bwilkerson: Is there a reason why you're not using `ResourceProvider` for this purpose?

angular_analyzer_plugin/test/plugin_test.dart
```diff
@@ -17,18 +22,34 @@ void main() {
 class PluginIntegrationTest {
   AngularAnalyzerPlugin plugin;
   ResourceProvider resourceProvider;
+  FileService fileService;
+  protocol.ContextRoot root;
 
   void setUp() {
     resourceProvider = PhysicalResourceProvider.INSTANCE;
```
bwilkerson: For testing, we usually create a `MemoryResourceProvider`. It allows us to create a file system in memory so we don't need to actually create and destroy files on disk, which makes the tests run much faster.

angular_analyzer_plugin/lib/src/file_service.dart
MichaelRFairhurst: Nope! Will change.

zoechi:
I'm wondering if

```yml
angular:
  enabled: true
```
is the best idea

The analyzer complains that `angular` is not a valid key

Perhaps something like

```yml
plugins: 
  angular:
    enabled: true
```
can be used where `plugins` is made known to analyzer so that it doesn't complain and ignores child values.
If you have another strategy in mind, please just ignore my comment.

bwilkerson:
I like the idea of putting all of the plug-in specific options under a top-level `plugins` key. It will reduce the change of name conflicts later.

MichaelRFairhurst:
Agreed, the reasoning is solid. Will make the change.

>

PR #396, opened 2017-07-19
Fix #394 plain boolean attributes resolved differently now

### LINKED ISSUE #394: Binding boolean inputs without brackets has new behavior
So

```
  <my-component someInput>
```

used to always be equivalent to

```
  <my-component [someInput]="''">
```

now, if `someInput` is a `bool`, then its equivalent to

```
  <my-component [someInput]="true">
```

For this reason, we will incorrectly flag the original example when its a boolean right now.

MichaelRFairhurst:


MichaelRFairhurst:
Adding Matan to reviewers since Harry is OOO

PR #398, opened 2017-07-25
Salt hashes for version updates of analysis/summary file formats

MichaelRFairhurst:
Track file hashes the same for html/non-html, do all hashing through
FileTracker which has a salt it applies to everything. Tests that dart
signatures, html signatures, content signatures, and unit element
signatures are salted. Refactor angular driver to use file tracker's
APIs all the time.

Also looks like file tracker tests weren't running before via
test_all.dart; fixed. And it had a test that was failing (presumably,
since it wasn't running!). Fixed that too.

Salt is currently at 1 since we're on version 1. Well, two, but the
presence of the salt distinguishes that much by itself.

### DIFF
--- a/angular_analyzer_plugin/lib/src/file_tracker.dart
+++ b/angular_analyzer_plugin/lib/src/file_tracker.dart
@@ -8,6 +8,8 @@ abstract class FileHasher {
 }
 
 class FileTracker {
+  static const int salt = 1;
+
   final FileHasher _fileHasher;
 
   FileTracker(this._fileHasher);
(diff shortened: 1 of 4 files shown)


MichaelRFairhurst:
Switching reviewers since Ming is OOO

angular_analyzer_plugin/lib/src/file_tracker.dart
```diff
@@ -82,24 +85,39 @@ class FileTracker {
...
     return signature;
   }
+
+  ApiSignature getContentSignature(String path) {
+    if (contentHashes[path] == null) {
+      rehashContents(path);
+    }
+    return new ApiSignature()
+      ..addInt(salt)
+      ..addBytes(_getContentHash(path));
```
scheglov: This (and similar code) is going to be expensive. MD5 is not cheap. If you know that content for a file has not changed, it is worth to cache results of `ApiSignature.toHex()` here.

PR #403, opened 2017-07-31
Add VS Code to editor list.

DanTup:
I got validation messages, but wasn't able to see any code completion etc. (though I don't have any real projects so don't totally understand what should/shouldn't work).


MichaelRFairhurst:
Looks great, the link to the ticket is great. I need to do something like that for vim!

My understanding from @lorenvs is that code completion worked. We can probably create a setup to test navigation, which only works under hard-to-recreate conditions.

I'll merge this and can update it as needed!

DanTup:
I don't have any Angular projects to give this a good test on, but if @LorenVS is using it I'm sure he can update the info here if required.

Thanks!

PR #406, opened 2017-08-02
Don't set completion offset/length completing outside templates

MichaelRFairhurst:


angular_analyzer_plugin/test/completion_contributor_test.dart
```diff
@@ -21,6 +21,18 @@ class DartCompletionContributorTest extends AbstractCompletionContributorTest {
...
   // ignore: non_constant_identifier_names
+  Future test_completeOutsideTemplateOK() async {
+    addTestSource('''
+import 'package:angular2/angular2.dart';
+class MyComp {
+  String text = ^;
+}
+    ''');
+
+    await computeSuggestions();
```
mk13: Maybe add test for completion offset and length? I believe they're at a default value if they're not overriden during the completion process.

PR #411, opened 2017-08-07
Do not depend on front_end directly.

stereotype441:
The front_end API is not stable.  Instead depend on the parts of
front_end re-exported by analyzer.

### DIFF
--- a/angular_analyzer_plugin/pubspec.yaml
+++ b/angular_analyzer_plugin/pubspec.yaml
@@ -18,6 +18,8 @@ dependency_overrides:
     path: ../deps/sdk/pkg/analyzer
   analyzer_plugin:
     path: ../deps/sdk/pkg/analyzer_plugin
+  front_end:
+    path: ../deps/sdk/pkg/front_end
 dev_dependencies:
   test_reflective_loader: '^0.1.0'
   test: '^0.12.20'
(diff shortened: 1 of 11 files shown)


angular_analyzer_plugin/tools/analyzer_plugin/pubspec.yaml.defaults
```diff
@@ -11,8 +11,6 @@ dependency_overrides:
     path: ABSOLUTE_PATH_TO_REPO/angular_analyzer_plugin
   analyzer:
     path: ABSOLUTE_PATH_TO_REPO/deps/sdk/pkg/analyzer
-  front_end:
```
MichaelRFairhurst: We still needs this, unfortunately, because it requires all transitive sdk dependencies (that's why we import kernel for instance) or we get version errors

angular_analyzer_plugin/pubspec.yaml
```diff
@@ -18,8 +18,6 @@ dependency_overrides:
     path: ../deps/sdk/pkg/analyzer
   analyzer_plugin:
     path: ../deps/sdk/pkg/analyzer_plugin
-  front_end:
```
MichaelRFairhurst: I think we need this too for the same reason as the other one...

angular_analyzer_plugin/tools/analyzer_plugin/pubspec.yaml.defaults
stereotype441: Ok, done.

angular_analyzer_plugin/pubspec.yaml
stereotype441: Done.

PR #413, opened 2017-08-08
Fix rejection of innerHTML as an input due to missing dart:html annotation

MichaelRFairhurst:
The public setter (as well as the private member, which realistically
means nothing for most people) should be annotated, but it isn't. As
such, we can't get the DomName() for innerHTML as such (capitalized)
and then apply angular's map to it.

To ensure that we still _suggest_ `innerHtml`, the angular way, while
accepting `innerHTML`, the DOM way, reuse all existing methodologies
after doing a little switcharoo up front.

Luckily this is harmless for in the future when that annotation gets
fixed, but I hear it may be months and months away.

### DIFF
--- a/angular_analyzer_plugin/lib/src/standard_components.dart
+++ b/angular_analyzer_plugin/lib/src/standard_components.dart
@@ -142,6 +142,7 @@ class BuildStandardHtmlComponentsVisitor extends RecursiveAstVisitor {
   }
 
   /// dart:html is missing an annotation to fix this casing. Compensate.
+  /// TODO(mfairhurst) remove this fix once dart:html is fixed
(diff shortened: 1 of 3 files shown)


angular_analyzer_plugin/lib/src/standard_components.dart
```diff
@@ -141,9 +141,12 @@ class BuildStandardHtmlComponentsVisitor extends RecursiveAstVisitor {
         isHtml: true);
   }
 
+  /// dart:html is missing an annotation to fix this casing. Compensate.
```
mk13: Might be nice to add a TODO here as a quick way to find when to revert the compensation.

MichaelRFairhurst: Ah definitely!

PR #416, opened 2017-08-09
Fix some hints being generated by analyzer

bwilkerson:


playground/lib/overview_component.dart
```diff
@@ -2,8 +2,6 @@ import 'package:angular2/angular2.dart';
 import 'counter_component.dart';
 import 'bubbled_directive.dart';
 
-import 'package:angular_analyzer_plugin/enable.dart';
```
MichaelRFairhurst: This will stop the plugin from being loaded in our playground until we have the new loading system in place...is that ready yet?

bwilkerson: I'm not sure, so I'll back that change out for now.

PR #418, opened 2017-08-21
Make inheritance test ordering safe

MichaelRFairhurst:
note: the build is expected to have one test failure from https://github.com/dart-lang/sdk/issues/30508

### DIFF
--- a/angular_analyzer_plugin/test/angular_driver_test.dart
+++ b/angular_analyzer_plugin/test/angular_driver_test.dart
@@ -3410,22 +3410,23 @@ class FinalComponent
     await getDirectives(source);
     final component =
         directives.firstWhere((d) => d.classElement.name == 'FinalComponent');
-    final compInputs = component.inputs;
+    final compInputs = component.inputs.fold(<String, InputElement>{},
+        (acc, input) => {input.name: input}..addAll(acc));
     expect(compInputs, hasLength(4));
     {
-      final input = compInputs[0];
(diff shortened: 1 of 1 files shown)


angular_analyzer_plugin/test/angular_driver_test.dart
```diff
@@ -3410,22 +3410,23 @@ class FinalComponent
     await getDirectives(source);
     final component =
         directives.firstWhere((d) => d.classElement.name == 'FinalComponent');
-    final compInputs = component.inputs;
+    final compInputs = component.inputs.fold(<String, InputElement>{},
+        (acc, input) => {input.name: input}..addAll(acc));
```
bwilkerson: If you're expecting to add more tests related to each of the inputs in the future, then this is fine, but if not I might suggest the slightly more efficient and shorter
```
final names = component.inputs.map((component) => component.name);
expect(names, unorderedEquals(['mixin1Input', 'mixin2Input', 'interface1Input', 'interface2Input']));
```
If you do keep this, you could change the expectations below to
```
expect(input, isNotNull);
```

MichaelRFairhurst: Oh, great, yes. Much prefer the unorderedEquals approach. Likely any additional assertions I add will be from adding new test cases rather than expanding this one...it was a possibility, but not worth skipping your (much cleaner, concise) approach.

PR #421, opened 2017-08-29
Handle major Angular security contexts; script & style pending.

MichaelRFairhurst:
For an input with a security context, allow binding to that security
context's safe type instead of the input type. For security contexts
with no sanitization available (resource urls ie iframe.src, scripts),
report an error if bound with anything but the safe type.

Style sanitization started but not completed. Looks like we don't allow
binding to [style] at all currently. Need to look into why its not part
of dart:html and what other exceptions we need to handle and how we
need to handle them. Security will likely "just work" then...some new
tests will need to be written and some comments in other tests will
need to be uncommented.

Script sanitization seems to be something entirely different?

Store security context classes on StandardAngular, and then use them to
create StandardHtml. That does mean we need to flip the order that we
get them. I also put in extra blocks to make sure that when our plugin
runs without any package:angular to resolve, it will not try to get
standard angular forever and ever by forcing 'no priority' as a driver.

### DIFF
--- a/angular_analyzer_plugin/lib/src/standard_components.dart
+++ b/angular_analyzer_plugin/lib/src/standard_components.dart
@@ -79,7 +79,7 @@ class SecuritySchema {
       SecurityContext scriptSecurityContext,
       SecurityContext styleSecurityContext,
       SecurityContext resourceUrlSecurityContext}) {
-    // This written to be easily synced to angular's security
+    // This is written to be easily synced to angular's security
     _registerSecuritySchema(
         htmlSecurityContext, ['iframe|srcdoc', '*|innerHTML', '*|outerHTML']);
     _registerSecuritySchema(styleSecurityContext, ['*|style']);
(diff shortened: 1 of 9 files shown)


angular_analyzer_plugin/lib/src/standard_components.dart
```diff
@@ -25,16 +26,122 @@ class StandardAngular {
...
+    }
+  }
+
+  SecuritySchema(
+      {SecurityContext htmlSecurityContext,
+      SecurityContext urlSecurityContext,
+      SecurityContext scriptSecurityContext,
+      SecurityContext styleSecurityContext,
+      SecurityContext resourceUrlSecurityContext}) {
+    // This written to be easily synced to angular's security
```
bwilkerson: nit: "This written" --> "This is written"

angular_analyzer_plugin/test/resolver_test.dart
```diff
@@ -475,6 +475,137 @@ class TestPanel {
...
+class TestPanel {
+  SafeHtml html;
+  SafeUrl url;
+  SafeStyle style;
+  SafeResourceUrl resourceUrl;
+}
+''');
+    final code = r"""
+<a [innerHtml]='html' [innerHTML]='html' [href]='url'></a>
+<iframe [src]='resourceUrl'></iframe><!-- TODO test [style]='style' -->
```
leonsenft: Have you filed issues for the TODOs to make sure they're not forgotten?

MichaelRFairhurst: Done!

angular_analyzer_plugin/lib/src/standard_components.dart
MichaelRFairhurst: Done! Thanks

PR #424, opened 2017-09-01
Reresolve inputs & outputs from parent classes to the child impls

MichaelRFairhurst:
This will support covariant/contravariant inputs/outputs, and also
means that navigation will navigate to the implementation that dart
itself will use when the input/output is bound.

Not yet removing input: or output: from @Directive. I did test that it
works, but don't have good infrastructure to make that an easy unit
test. From my tests, our code works whether the definition of Directive
accepts those fields or not, so we're backwards compatible right now
which is great.

Added tests that generic base components resolve the types properly.

Had to move some tests from testing first pass directive collection, to
directive linking. First pass does not include directive inheritance.

ContentChild(ren) coming soon.

### DIFF
--- a/angular_analyzer_plugin/lib/src/directive_linking.dart
+++ b/angular_analyzer_plugin/lib/src/directive_linking.dart
@@ -350,8 +350,8 @@ class InheritedMetadataLinker {
   }
 
   InputElement reresolveInput(InputElement input) {
-    final setter = directive.classElement.lookUpSetter(
-        input.setter.name.replaceAll('=', ''), directive.classElement.library);
+    final setter = directive.classElement
+        .lookUpSetter(input.setter.displayName, directive.classElement.library);
     if (setter == null) {
       // Happens when an interface with an input isn't implemented correctly.
       // This will be accompanied by a dart error, so we can just return the
--- a/angular_analyzer_plugin/lib/src/directive_linking.dart
+++ b/angular_analyzer_plugin/lib/src/directive_linking.dart
@@ -320,8 +320,15 @@ class ExportLinker {
 class InheritedMetadataLinker {
   AbstractDirective directive;
   FileDirectiveProvider _fileDirectiveProvider;
+  BindingTypeSynthesizer bindingSynthesizer;
 
-  InheritedMetadataLinker(this.directive, this._fileDirectiveProvider);
+  InheritedMetadataLinker(this.directive, this._fileDirectiveProvider)
+      : bindingSynthesizer = new BindingTypeSynthesizer(
+            directive.classElement,
+            directive.classElement.library.definingCompilationUnit.context
+                .typeProvider,
+            directive.classElement.library.definingCompilationUnit.context,
+            new ErrorReporter(new IgnoringErrorListener(), directive.source));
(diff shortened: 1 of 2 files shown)


angular_analyzer_plugin/lib/src/directive_linking.dart
```diff
@@ -335,12 +342,50 @@ class InheritedMetadataLinker {
...
+      directive.inputs.addAll(match.inputs.map(reresolveInput));
+      directive.outputs.addAll(match.outputs.map(reresolveOutput));
       directive.contentChildFields.addAll(match.contentChildFields);
       directive.contentChildrenFields.addAll(match.contentChildrenFields);
     }
   }
+
+  InputElement reresolveInput(InputElement input) {
+    final setter = directive.classElement.lookUpSetter(
+        input.setter.name.replaceAll('=', ''), directive.classElement.library);
```
leonsenft: You can use `input.setter.displayName` to get the property name without the `'='`.

angular_analyzer_plugin/lib/src/directive_linking.dart
```diff
@@ -335,12 +342,50 @@ class InheritedMetadataLinker {
...
+    if (setter == null) {
+      // Happens when an interface with an input isn't implemented correctly.
```
leonsenft: Nice comment, thanks for this. :smiley:

angular_analyzer_plugin/lib/src/directive_linking.dart
```diff
@@ -335,12 +342,50 @@ class InheritedMetadataLinker {
         continue;
       }
 
-      directive.inputs.addAll(match.inputs);
-      directive.outputs.addAll(match.outputs);
+      directive.inputs.addAll(match.inputs.map(reresolveInput));
```
leonsenft: Isn't this list going to end up with multiple instances of the same input if its overridden? Is de-duplication/overriding handled later?

MichaelRFairhurst: ah, yes, great question. If both are annotated with `@Input`, duplication does currently occur. I figure I will solve that once I tackle validating conflicting overrides

angular_analyzer_plugin/lib/src/directive_linking.dart
MichaelRFairhurst: Ooh! Definitely will use that, thanks!

angular_analyzer_plugin/lib/src/directive_linking.dart
leonsenft: Oh since this is just for covariance/contravariance it doesn't matter which input definition you read right? I'm guessing where you check for assignment this is handled correctly?

MichaelRFairhurst: yep!

I *think* the result of duplication would mean that you'd have multiple copies of the correctly configured input. So, you'd probably get two errors in place of 1 when you bind to it incorrectly.

It would be an issue where the two inputs have different names, but I know that's an area we're still not sure which direction to take.

PR #426, opened 2017-09-06
Support <audio> tags with tests (exceptional tag code).

MichaelRFairhurst:


angular_analyzer_plugin/test/mock_sdk.dart
```diff
@@ -443,6 +443,22 @@ class TemplateElement extends HtmlElement {
...
 }
+
+class AudioElement extends MediaElement {
+  factory AudioElement._([String src]) {
+    if (src != null) {
+      return AudioElement._create_1(src);
+    }
+    return AudioElement._create_2();
+  }
+  static AudioElement _create_1(src) => JS('AudioElement', 'new Audio(#)', src);
```
liuming0: Make single quote or double quote consistent in the file.

angular_analyzer_plugin/test/mock_sdk.dart
```diff
@@ -443,6 +443,22 @@ class TemplateElement extends HtmlElement {
...
+  static AudioElement _create_2() => JS('AudioElement', 'new Audio()');
+  AudioElement.created() : super.created();
+
+  factory AudioElement([String src]) => new AudioElement._(src);
+}
+
+class MediaElement extends Element {}
```
liuming0: Why a empty class?

angular_analyzer_plugin/test/mock_sdk.dart
MichaelRFairhurst: Ah, so this file is actually a copy of "dart:html", part of our "mock sdk."

We scan dart:html for this in order to get type information rather than a regular HTML schema.

Most other classes are trimmed down copies of their source. In this case, I copied the source of "AudioElement" exactly. Since that includes "extends MediaElement", I included a "MediaElement extends Element" class, but didn't fill it in since it doesn't (currently seem to) have any edge cases.

PR #428, opened 2017-09-06
Update readme with usage instructions/playground project for new loader

MichaelRFairhurst:
Now that plugin discovery has changed, update the playground project to
use that new discovery mode. And update the "how to hack on this
project" setup instructions as well.

Mark the deprecated build process as deprecated.

Add new install instructions which will become part of the default
bootstrap project/instructions some day.


README.md
```diff
@@ -10,22 +10,32 @@
...
+Using this strategy allows the dart ecosystem to discover which version of our plugin will work best with your version of angular, and therefore is recommended.
+
+Simply add to your [analysis_options.yaml file](https://www.dartlang.org/guides/language/analysis-options#the-analysis-options-file):
 
-Then run 
 ```
-./tools/get_deps.sh
-cd old_plugin_loader/bin
-./make_snapshot
+# forwards compatible
```
kevmoo: forwards vs backwards is confusing – help?

kevmoo: So I need both?

MichaelRFairhurst: Yep. Still need to support the plugin loader available within the g3 SDK, which is different than the new one which is the only one that works with pub.

In a few weeks, only the forward compatible version will be required.

Of course, I could have forked here or there, but this is for my own sanity. Should be short lived.

PR #430, opened 2017-09-08
Check reductions are only on keyup keydown

MichaelRFairhurst:
This is the broadest type of validation that we can do easily.
Shouldn't introduce any false errors.

Not as complete as validating the available keys and modifiers and
their ordering, but that is less important than warning people about
when reductions apply or don't apply at all.


angular_analyzer_plugin/lib/ast.dart
```diff
@@ -176,6 +176,10 @@ class StatementsBoundAttribute extends BoundAttributeInfo {
   @override
   String toString() => '(${super.toString()}, [$statements])';
 
+  int get reductionsOffset => nameOffset + name.length;
+  int get reductionsLength =>
+      reductions.isEmpty ? 0 : '.'.length + reductions.join('.').length;
```
mk13: Might be nice to add a test case for these two.

MichaelRFairhurst: Done! agreed.

PR #431, opened 2017-09-08
Integration fixes -- fix NPE and angular_ast versioning differences during testing

MichaelRFairhurst:


angular_analyzer_plugin/lib/src/navigation.dart
```diff
@@ -67,7 +67,8 @@ class AngularNavigation implements NavigationContributor {
       collector.addRegion(
           input.setterRange.offset,
           input.setterRange.length,
-          null, //protocol.convertElementKind(setter.kind),
+          protocol.ElementKind.UNKNOWN, // TODO create & use convertElement API
+          // protocol.convertElementKind(setter.kind),
```
bwilkerson: This is already implemented for you in analyzer_plugin/lib/utilities/analyzer_converter.dart

MichaelRFairhurst: Ah! I found it! I thought it was only implemented in `package:analysis_server/protocol`, and not  `package:analyzer_plugin/protocol`.

Done! Thanks!

PR #438, opened 2017-09-14
Fix tests running in checked mode, and make that default on travis

MichaelRFairhurst:
Had one type error, that had to be fixed.

Also was tripping an assertion in the analyzer's autocomplete that
expressions must have parents. The solution here is perhaps too hacky,
but perhaps not. Just wrap in a virtual parenthesized expression.
Doesn't seem like a problem to pass in null for the open and close
paren tokens.

### DIFF
--- a/angular_analyzer_plugin/lib/src/completion_request.dart
+++ b/angular_analyzer_plugin/lib/src/completion_request.dart
@@ -1,6 +1,7 @@
 import 'package:analyzer/error/error.dart';
 import 'package:analyzer/error/listener.dart';
 import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/standard_ast_factory.dart';
 import 'package:analyzer/file_system/file_system.dart';
 import 'package:analyzer_plugin/utilities/completion/completion_core.dart';
 import 'package:analyzer_plugin/src/utilities/completion/completion_target.dart';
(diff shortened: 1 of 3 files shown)


angular_analyzer_plugin/lib/src/completion_request.dart
```diff
@@ -63,6 +64,12 @@ class AngularCompletionRequest extends CompletionRequest {
       _dartSnippet = extractor.dartSnippet;
       _angularTarget = extractor.target;
       if (_dartSnippet != null) {
+        if (_dartSnippet is Expression) {
+          // wrap dart snippet in a ParenthesizedExpression, because the dart
+          // completion engine expects all expressions to have parents.
```
bwilkerson: Do you have an idea of how much work it would be to remove this expectation? Seems like it might be a better solution, but might not be practical at this point.

MichaelRFairhurst: It seems like maybe a pretty fundamental part of the autocompleter?

Here's the comment above the assert that we're tripping otherwise:

```
    // This should never happen; the containingNode will always be some node     
    // higher up in the parse tree, and the SimpleIdentifier will be the         
    // entity.      
```

So it seems like as long as it expects to find both a containingNode and a SimpleIdentifier, then all nodes should need a parent?

PR #440, opened 2017-09-15
Properly report a hint (not error) for overly complicated templates

MichaelRFairhurst:
Should have been a hint all along, changed that.

Also, seems like I wrote tests for this but just expected the wrong
error. Solved.

Aaaand the implementation is fixed too, that helps. Basically, since
the OffsettingConstantEvaluator doesn't actually know everything
required to actually resolve the string value, some valid string
templates are not actually strings, but rather, "unknown." Don't report
its not a string just because of this.

The fact that we don't get the actual value in this case doesn't
matter; we won't analyze it.


angular_analyzer_plugin/lib/errors.dart
```diff
@@ -457,3 +450,25 @@ class AngularWarningCode extends ErrorCode {
   @override
   ErrorType get type => ErrorType.STATIC_WARNING;
 }
+
+class AngularHintCode extends AngularWarningCode {
+  /// An error code indicating that an output-bound statement
+  /// must be an [ExpressionStatement].
```
mk13: I think above is a better error message than what is currently set to.

MichaelRFairhurst: Ah! good catch.

PR #444, opened 2017-10-12
Refactor model to support functional directives

MichaelRFairhurst:
Doesn't actually introduce any new behavior, yet. But the ability to add FunctionalDirectives -- in theory -- will work from here.

### DIFF
--- a/angular_analyzer_plugin/lib/src/model.dart
+++ b/angular_analyzer_plugin/lib/src/model.dart
@@ -15,44 +15,123 @@ import 'package:angular_analyzer_plugin/errors.dart';
 
 /// Might be a directive, or a component, or neither. It might simply have
 /// annotated @Inputs, @Outputs() intended to be inherited.
-class AngularAnnotatedClass {
+class AngularAnnotatedClass extends AngularTopLevel {
   /// The [ClassElement] this annotation is associated with.
   final dart.ClassElement classElement;
-  final List<InputElement> inputs;
-  final List<OutputElement> outputs;
+
+  AngularAnnotatedClass(this.classElement,
+      {List<InputElement> inputs,
+      List<OutputElement> outputs,
+      this.contentChildFields,
+      this.contentChildrenFields})
+      : super(inputs: inputs, outputs: outputs);
+
+  /// The source that contains this directive.
+  @override
+  Source get source => classElement.source;
+
+  @override
+  String toString() => '$runtimeType(${classElement.displayName} '
+      'inputs=$inputs '
+      'outputs=$outputs '
+      'attributes=$attributes)';
+
+  @override
+  bool operator ==(Object other) =>
+      other is AngularAnnotatedClass && other.classElement == classElement;
+
+  @override
+  int get hashCode => classElement.hashCode;
+
+  @override
+  bool get isHtml => false;
+
+  // See [contentChildrenFields]. These are the linked versions.
+  @override
+  final contentChilds = <ContentChild>[];
+  @override
+  final contentChildren = <ContentChild>[];
 
   /// Which fields have been marked `@ContentChild`, and the range of the type
   /// argument. The element model contains the rest. This should be stored in the
   /// summary, so that at link time we can report errors discovered in the model
   /// against the range we saw it the AST.
+  @override
   List<ContentChildField> contentChildrenFields;
+  @override
   List<ContentChildField> contentChildFields;
-
-  AngularAnnotatedClass(this.classElement, this.inputs, this.outputs,
-      this.contentChildFields, this.contentChildrenFields);
 }
 
-/// An abstract model of an Angular directive.
-abstract class AbstractDirective extends AngularAnnotatedClass {
-  final AngularElement exportAs;
+/// A functionoal directive is applied when the directive is linked, but does
+/// nothing later in the program. Thus it cannot have inputs, outputs, etc. But
+/// for the sake of clean code, those methods are implemented to return null,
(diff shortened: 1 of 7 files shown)


angular_analyzer_plugin/lib/src/model.dart
```diff
@@ -15,70 +15,178 @@ import 'package:angular_analyzer_plugin/errors.dart';
...
   List<ContentChildField> contentChildFields;
-
-  AngularAnnotatedClass(this.classElement, this.inputs, this.outputs,
-      this.contentChildFields, this.contentChildrenFields);
 }
 
-/// An abstract model of an Angular directive.
-abstract class AbstractDirective extends AngularAnnotatedClass {
-  final AngularElement exportAs;
+/// A functionoal directive is applied when the directive is linked, but does
```
mk13: Typo.

angular_analyzer_plugin/lib/src/model.dart
```diff
@@ -15,70 +15,178 @@ import 'package:angular_analyzer_plugin/errors.dart';
...
-  bool operator ==(Object other) =>
-      other is AbstractDirective && other.classElement == classElement;
+  String get name;
 
-  @override
-  int get hashCode => classElement.hashCode;
+  List<ElementNameSelector> get elementTags;
+}
+
+/// An abstract model of an Angular top level construct. This may be a
```
mk13: Nit: The code base  hasn't been doing this, but I dunno if you want to start: https://www.dartlang.org/guides/language/effective-dart/documentation#do-make-the-first-sentence-a-standalone-paragraph

or just keep it as it is for consistency. Your call.

angular_analyzer_plugin/lib/src/model.dart
```diff
@@ -15,70 +15,178 @@ import 'package:angular_analyzer_plugin/errors.dart';
...
+  final attributes = <AngularElement>[];
+
+  bool get isHtml;
+  Source get source;
+
+  /// Its very hard to tell which directives are meant to be used with a *star.
+  /// However, any directives which have a `TemplateRef` as a constructor
+  /// parameter are almost certainly meant to be used with one. We use this for
+  /// whatever validation we can, and autocomplete suggestions.
+  bool looksLikeTemplate = false;
```
mk13: Nit: This seems more like an internal comment; prefer using '//' instead '///' which is for documentation.

angular_analyzer_plugin/lib/src/model.dart
```diff
@@ -15,70 +15,178 @@ import 'package:angular_analyzer_plugin/errors.dart';
...
+
+  final List<InputElement> inputs;
+  final List<OutputElement> outputs;
+
+  // See [AngularAnnotatedClassMembers.contentChildren]
```
mk13: Seems more like a '///'.

angular_analyzer_plugin/lib/src/model.dart
```diff
@@ -15,70 +15,178 @@ import 'package:angular_analyzer_plugin/errors.dart';
...
+  List<ContentChild> get contentChilds;
+  List<ContentChild> get contentChildren;
+
+  // See [AngularAnnotatedClassMembers.contentChildrenFields]
```
mk13: Same here.

angular_analyzer_plugin/test/angular_driver_test.dart
```diff
@@ -3689,6 +3678,39 @@ class MyComponent extends Generic<String> {
...
+@Directive(selector: '[attr]')
+class PartDirective {}
+        ''');
+    await getDirectives(partSource);
+    errorListener.assertNoErrors();
+    await getDirectives(source);
+    final Component component = directives.single;
+    final childDirectives = component.view.directives;
+    expect(childDirectives, hasLength(1));
+    final directive = childDirectives.first;
```
mk13: I found out recently that if you do: childDirectives.single, it does length checking and gets you the first element.

angular_analyzer_plugin/lib/src/model.dart
MichaelRFairhurst: Fixed!

angular_analyzer_plugin/lib/src/model.dart
MichaelRFairhurst: I like it. Fixed!

angular_analyzer_plugin/lib/src/model.dart
MichaelRFairhurst: Fixed!

angular_analyzer_plugin/test/angular_driver_test.dart
MichaelRFairhurst: Ah, awesome. Also this was a test that slipped in, I had been trying to repro a bug. Will remove it, but I'll update that wherever I come across it -- much better pattern.

angular_analyzer_plugin/lib/src/model.dart
MichaelRFairhurst: Fixed!

angular_analyzer_plugin/lib/src/model.dart
MichaelRFairhurst: I think I'll leave this one here....I agree that it seems more of an internal comment...but I'd rather treat this as an external comment just in case this becomes part of external facing stuff later. I'd be curious to hear if you disagree in more detail...I can definitely fix it later in a different PR based on what we hash out.

PR #452, opened 2017-11-14
Allow directives: VARIABLE, using a ResolutionStrategy of list/const.

MichaelRFairhurst:
For list values we are able to report offsets of individual items, and
for const values we are not. Resolution is very different in this
respect and APIs used.

Track the angular standard classes which contain directives (view,
component), and use that resolution data to find annotations for the
case of const resolution -- this can be used in other cases too, like
pipes and selectors etc. It can also likely be used for optimizing the
current behavior.

New summary format to differentiate const/list strategy.

Tests that the `directives: VARIABLE` (or other const expression)
strategy works correctly.

### DIFF
--- a/angular_analyzer_plugin/lib/src/model.dart
+++ b/angular_analyzer_plugin/lib/src/model.dart
@@ -636,7 +636,6 @@ class UseConstValueStrategy implements DirectivesStrategy {
 
   /// Check if an element is a View or Component
   bool _isViewOrComponent(dart.Element element) =>
-      element != null &&
       element is dart.ClassElement &&
       (element.type.isSubtypeOf(standardAngular.view.type) ||
           element.type.isSubtypeOf(standardAngular.component.type));
@@ -665,7 +664,6 @@ class View implements AnalysisTarget {
   final Component component;
   final List<AbstractDirective> directives;
   final List<Pipe> pipes;
-  //final List<DirectiveReference> directiveReferences;
   final DirectivesStrategy directivesStrategy;
   final List<PipeReference> pipeReferences;
   final String templateText;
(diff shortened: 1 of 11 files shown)


angular_analyzer_plugin/lib/src/model.dart
```diff
@@ -594,6 +595,68 @@ class PipeReference {
...
+              .where((m) => _isViewOrComponent(m.element?.enclosingElement))
+              .map((m) => _getDirectives(m.computeConstantValue()))
+              // TODO(mfairhurst): report error for double definition
+              .firstWhere((directives) => !(directives?.isNull ?? true),
+                  orElse: () => null),
+          sourceRange);
+
+  /// Check if an element is a View or Component
+  bool _isViewOrComponent(dart.Element element) =>
+      element != null &&
```
scheglov: This check is not necessary, `value is T` is `false` for `null`.

angular_analyzer_plugin/lib/src/model.dart
```diff
@@ -602,7 +665,8 @@ class View implements AnalysisTarget {
   final Component component;
   final List<AbstractDirective> directives;
   final List<Pipe> pipes;
-  final List<DirectiveReference> directiveReferences;
+  //final List<DirectiveReference> directiveReferences;
```
scheglov: Do you still need this line?

MichaelRFairhurst: Definitely not, thanks!

angular_analyzer_plugin/lib/src/model.dart
MichaelRFairhurst: You're right. Hmm, I remember adding this because of a bunch of fuzz failures. I can't reproduce them, though, with or without this line removed. Must have solved it in a different way or something.

PR #456, opened 2017-11-20
Publish 0.0.11

MichaelRFairhurst:
Many many bugfixes since, even some in angular_ast. Definitely worth
publishing now.

Instructions for loading an exact version via the self-loader method,
which is updated here to load itself at the same version (0.0.11)


daniel-v:
Any ETA for this release?

MichaelRFairhurst:
This may have to wait until next Monday, tomorrow is Thanksgiving here and then we have Friday off as well. Most of my potential reviewers are already on vacation and those who aren't are mostly really busy.

I'm working on it though! Want to get it to you ASAP

PR #457, opened 2017-11-20
Support new plugin options, track custom_tag_names. Drop NoopDriver.

MichaelRFairhurst:
The server is now smart enough to not load the plugin when it isn't
enabled, so don't worry about creating a NoopDriver in that case any
more.

Since the plugin may be loaded either direct from angular, or via the
self-loader system, accept looking up options from either angular: or
angular_analyzer_plugin: via the analysis_options.yaml.

Try to be somewhat smart (at least, deliberate and predictable) when
both options exist. If not enabled, don't consider those options.

Doesn't yet
- affect summary signatures
- affect analysis
- report errors in the config


README.md
```diff
@@ -19,16 +19,10 @@ Using this strategy allows the dart ecosystem to discover which version of our p
...
     angular:
       enabled: true
-
-# backwards compatible (you also need this, for the moment)
-plugins:
-  angular:
-    enabled: true
 ```
 
 Eventually, only the forwards compatible section will be required.
```
bwilkerson: Remove this line too.

MichaelRFairhurst: Good catch! Done

PR #458, opened 2017-11-20
Incorporate custom_tag_names into unit signatures for summaries.

MichaelRFairhurst:
No analysis yet dependent on them, but this ensures the analysis won't
be stale once that analysis is added.

Also change 'unknownTagNames' to 'customTagNames' to reflect the
setting name in the yaml file, 'custom_tag_names' which is the better
term for what it does.

Comments about how summaries will be affected by future analysis
options and making sure they are in the right locations/are adequately
differentiable.

### DIFF
--- a/angular_analyzer_plugin/lib/src/file_tracker.dart
+++ b/angular_analyzer_plugin/lib/src/file_tracker.dart
@@ -1,6 +1,7 @@
 import 'dart:collection';
 
 import 'package:analyzer/src/summary/api_signature.dart';
+import 'package:angular_analyzer_plugin/src/options.dart';
 
 abstract class FileHasher {
   ApiSignature getContentHash(String path);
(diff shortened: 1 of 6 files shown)


angular_analyzer_plugin/lib/src/file_tracker.dart
```diff
@@ -11,8 +12,9 @@ class FileTracker {
   static const int salt = 4;
 
   final FileHasher _fileHasher;
+  final AngularOptions _options;
 
-  FileTracker(this._fileHasher);
+  FileTracker(this._fileHasher, this._options);
```
scheglov: This might be too much info for this class.
In a similar situation we pass just a "salt" into the `FileSystemState` in analyzer, which includes all options related bits.

MichaelRFairhurst: I do agree. I like that I can salt the different types of summaries differently, but that doesn't require passing in the whole options.

I'll submit for now since its a bit time sensitive, but make a ticket to change it to passing in the salts.

PR #459, opened 2017-11-20
Suppress many (but not all) errors when they occur on custom tags.

MichaelRFairhurst:
Inputs and outputs may be because of web components, ignore those
warnings on a tag that was entered as a custom tag.

Ensure we stil get all the rest.

### DIFF
--- a/angular_analyzer_plugin/lib/src/resolver.dart
+++ b/angular_analyzer_plugin/lib/src/resolver.dart
@@ -16,6 +16,7 @@ import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/src/generated/source.dart';
 import 'package:angular_analyzer_plugin/src/facade/exports_library_element.dart';
 import 'package:angular_analyzer_plugin/src/model.dart';
+import 'package:angular_analyzer_plugin/src/options.dart';
 import 'package:angular_analyzer_plugin/src/selector.dart';
 import 'package:angular_analyzer_plugin/src/standard_components.dart';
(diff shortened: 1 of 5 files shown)


angular_analyzer_plugin/lib/src/resolver.dart
```diff
@@ -1027,7 +1041,7 @@ class ComponentContentResolver extends AngularAstVisitor {
   }
 
   /// Check whether the given [name] is a standard HTML5 tag name.
-  static bool _isStandardTagName(String name) {
+  bool _isStandardTagName(String name) {
     // ignore: parameter_assignments
     name = name.toLowerCase();
     return !name.contains('-') || name == 'ng-content';
```
jonahwilliams: I'm not sure whether it makes sense to include in the AAP, but there are additional restrictions around custom element names that might make sense here.

http://w3c.github.io/webcomponents/spec/custom/#custom-elements-core-concepts

There are also some svg tags which contain '-' like font-face-uri that you might want to consider.

MichaelRFairhurst: Yeah...this is actually a really terrible function. I have an open ticket for it, #248, which we might be ready to revisit now that you mention it.

Other problems here include that some people make custom tags without dashes, and they will get no validation on its spelling when they do!

jonahwilliams: I _think_ that polymer custom elements have the same restrictions as the browser custom elements, so you should only encounter `-` in custom elements

PR #460, opened 2017-11-21
Change file-prefixed errors for "weird" include paths to be clearer

MichaelRFairhurst:
Requires more summary data, so also have to tick the summary salt.

### DIFF
--- a/angular_analyzer_plugin/lib/src/from_file_prefixed_error.dart
+++ b/angular_analyzer_plugin/lib/src/from_file_prefixed_error.dart
@@ -1,30 +1,32 @@
 import 'package:analyzer/error/error.dart';
 import 'package:analyzer/src/generated/source.dart';
 
-// A wrapper around AnalysisError which also links back to a "from" file for
-// context.
+// A wrapper around AnalysisError which also links back to a "from" file and
+// classname for context.
 //
 // Is a wrapper, not just an extension, so that it can have a different hashCode
 // than the error without a "from" path, in the case that a file is included
 // both sanely and strangely (which is common: prod and test).
 class FromFilePrefixedError implements AnalysisError {
   final String fromSourcePath;
+  final String classname;
   final String originalMessage;
   final AnalysisError originalError;
   String _message;
 
-  FromFilePrefixedError(Source fromSource, AnalysisError originalError)
+  FromFilePrefixedError(
+      Source fromSource, this.classname, AnalysisError originalError)
       : originalMessage = originalError.message,
         fromSourcePath = fromSource.fullName,
         originalError = originalError {
-    _message = "$originalMessage (from $fromSourcePath)";
+    _message = "In $classname: $originalMessage (from $fromSourcePath)";
   }
 
   FromFilePrefixedError.fromPath(
-      this.fromSourcePath, AnalysisError originalError)
+      this.fromSourcePath, this.classname, AnalysisError originalError)
       : originalMessage = originalError.message,
         originalError = originalError {
-    _message = "$originalMessage (from $fromSourcePath)";
+    _message = "In $classname: $originalMessage (from $fromSourcePath)";
   }
 
   @override
--- a/angular_analyzer_plugin/lib/src/from_file_prefixed_error.dart
+++ b/angular_analyzer_plugin/lib/src/from_file_prefixed_error.dart
@@ -15,17 +15,17 @@ class FromFilePrefixedError implements AnalysisError {
   String _message;
 
   FromFilePrefixedError(
-      Source fromSource, this.classname, AnalysisError originalError)
-      : originalMessage = originalError.message,
-        fromSourcePath = fromSource.fullName,
-        originalError = originalError {
-    _message = "In $classname: $originalMessage (from $fromSourcePath)";
-  }
+      Source fromSource, String classname, AnalysisError originalError)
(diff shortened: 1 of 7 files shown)


angular_analyzer_plugin/lib/src/from_file_prefixed_error.dart
```diff
@@ -1,30 +1,32 @@
...
   String _message;
 
-  FromFilePrefixedError(Source fromSource, AnalysisError originalError)
+  FromFilePrefixedError(
+      Source fromSource, this.classname, AnalysisError originalError)
       : originalMessage = originalError.message,
         fromSourcePath = fromSource.fullName,
         originalError = originalError {
-    _message = "$originalMessage (from $fromSourcePath)";
+    _message = "In $classname: $originalMessage (from $fromSourcePath)";
```
stereotype441: It looks from the code below like `classname` will default to the empty string if not specified.  This will produce a weird looking message like `In : <originalMessage> (from <fromSourcePath>)`.  How about if we change this to something like:

```dart
_message = (classname.isEmpty ? "" : "In $classname: ") + "$originalMessage (from $fromSourcePath)";
```

angular_analyzer_plugin/lib/src/from_file_prefixed_error.dart
```diff
@@ -1,30 +1,32 @@
...
 // both sanely and strangely (which is common: prod and test).
 class FromFilePrefixedError implements AnalysisError {
   final String fromSourcePath;
+  final String classname;
   final String originalMessage;
   final AnalysisError originalError;
   String _message;
 
-  FromFilePrefixedError(Source fromSource, AnalysisError originalError)
+  FromFilePrefixedError(
```
stereotype441: Nit: we could share more code if we implemented this constructor as:

```dart
FromFilePrefixedError(
    Source fromSource, String className, AnalysisError originalError)
  : this.fromPath(fromSource.fullName, className, originalError);
```

angular_analyzer_plugin/lib/src/from_file_prefixed_error.dart
```diff
@@ -1,30 +1,32 @@
...
+    _message = "In $classname: $originalMessage (from $fromSourcePath)";
   }
 
   FromFilePrefixedError.fromPath(
-      this.fromSourcePath, AnalysisError originalError)
+      this.fromSourcePath, this.classname, AnalysisError originalError)
       : originalMessage = originalError.message,
         originalError = originalError {
-    _message = "$originalMessage (from $fromSourcePath)";
+    _message = "In $classname: $originalMessage (from $fromSourcePath)";
```
stereotype441: The same change needs to be made here (or share the code as I suggested earlier)

angular_analyzer_plugin/lib/src/summary/idl.dart
```diff
@@ -147,6 +147,8 @@ abstract class SummarizedAnalysisErrorFromPath extends base.SummaryClass {
   @Id(0)
   String get path;
   @Id(1)
+  String get classname;
+  @Id(2)
```
stereotype441: Did you mean to change the Id of `originalError` from 1 to 2?  It would be better for backwards/forwards compatibility to do:

```dart
abstract class SummarizedAnalysisErrorFromPath extends base.SummaryClass {
  @Id(0)
  String get path;
  @Id(2)
  String get classname;
  @Id(1)
  SummarizedAnalysisError get originalError;
}
```

angular_analyzer_plugin/lib/src/from_file_prefixed_error.dart
MichaelRFairhurst: Absolutely, great suggestion

angular_analyzer_plugin/lib/src/from_file_prefixed_error.dart
MichaelRFairhurst: how about assert(classname != null); assert(className.isNotEmpty) ?

angular_analyzer_plugin/lib/src/from_file_prefixed_error.dart
MichaelRFairhurst: Refactored to a private method so they can share

angular_analyzer_plugin/lib/src/summary/idl.dart
MichaelRFairhurst: I decided to not worry about backwards compatibility, this error is confusing so I ticked the summary salt. The reanalysis cost is unfortunate, but, the clearer errors that will suddenly come from this almost certainly be worth it

angular_analyzer_plugin/lib/src/from_file_prefixed_error.dart
stereotype441: Works for me

angular_analyzer_plugin/lib/src/summary/idl.dart
stereotype441: Fair enough

angular_analyzer_plugin/lib/src/from_file_prefixed_error.dart
MichaelRFairhurst: Ah right, the constructor fix also reduced the code duplication. That's done now.

PR #466, opened 2017-11-30
Add new options for ContentChild(ren) in prep for deprecating ElementRef

MichaelRFairhurst:
Accept (for the moment) ElementRef, Element, and HtmlElement (the
latter two being from dart:html).

Update most tests specific to ElementRef to have a copy that uses
Element. Tests that use ElementRef to check general
transclusion/content child/content children behavior which use
ElementRef have been changed to use Element.

We currently don't differentiate SVG and HTML, so we accept either type
for either case at the moment.

### DIFF
--- a/angular_analyzer_plugin/test/resolver_test.dart
+++ b/angular_analyzer_plugin/test/resolver_test.dart
@@ -3911,6 +3911,30 @@ class TranscludeSome {
 
   Future
       // ignore: non_constant_identifier_names
+      test_resolveTemplate_provideContentNotMatchingSelectorsButMatchesContentChildHtmlElement() async {
+    _addDartSource(r'''
+import 'dart:html';
+@Component(selector: 'test-panel', templateUrl: 'test_panel.html',
+    directives: const [TranscludeSome])
+class TestPanel {
+}
+@Component(selector: 'transclude-some',
+    template: '<ng-content select="transclude-me"></ng-content>')
+class TranscludeSome {
+  @ContentChild(HtmlElement)
+  HtmlElement foo;
+}
+''');
+    final code = r"""
+<transclude-some><div></div></transclude-some>
+    """;
+    _addHtmlSource(code);
+    await _resolveSingleTemplate(dartSource);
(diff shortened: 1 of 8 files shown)


angular_analyzer_plugin/test/resolver_test.dart
```diff
@@ -3957,16 +3981,19 @@ class ContentChildComponent {
...
+import 'dart:html';
 @Component(selector: 'test-panel', templateUrl: 'test_panel.html',
     directives: const [TranscludeNone])
 class TestPanel {
 }
 @Component(selector: 'transclude-none', template: '')
 class TranscludeNone {
   @ContentChild('contentChild')
-  ElementRef foo;
+  Element foo;
```
leonsenft: Querying for an `Element` or `HtmlElement` by reference without a `read` token isn't supported yet. The issue of inferring the read type came up in https://github.com/dart-lang/angular/issues/537, but it's not yet resolved. This will currently fail at run-time as it will attempt to assign an `ElementRef` to `foo`.

MichaelRFairhurst: Ah, fantastic catch!! Thank you, I had no idea. Will fix.

MichaelRFairhurst:
> I didn't see any tests for HtmlElement, only Element. Should we add some?

Ah, I suppose I should, yes!

MichaelRFairhurst:
@leonsenft  PTAL

MichaelRFairhurst:
Yep! I have

```
  @ContentChild('el', read: HtmlElement)                                         
  Element elemNotHtmlElem;  
```

as well as

```
  @ContentChild('el', read: Element)                                             
  HtmlElement htmlElemNotElem;  
```

With assertions for errors in the latter and no errors in the former.

MichaelRFairhurst:
Working on the travis failure now. I essentially need to have some branches (such as this one) test against published versions of the analyzer, and other versions (like what we pull into google) to test against the latest versions of the analyzer.

leonsenft:
LGTM pending Travis fixes.

PR #469, opened 2017-12-07
Test travis with pub

MichaelRFairhurst:


MichaelRFairhurst:
This branch has history in it it should not have. Remaking.

PR #473, opened 2017-12-07
Fix travis branch detection; git branch doesn't seem to work.

MichaelRFairhurst:
Use the travis environment variable instead -- but still fall back to
git branch for local runs.

This only affects branches running with SDK_AT_HEAD.

### DIFF
--- a/tools/travis.sh
+++ b/tools/travis.sh
@@ -3,6 +3,11 @@
 # Fast fail the script on failures.
 set -e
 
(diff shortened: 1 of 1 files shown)


MichaelRFairhurst:
Oops, I shouldn't have used the new name here, because it will fail until #472 is landed, which I'm not going to land until its travis build passes, which will fail until this is landed......

PR #475, opened 2017-12-08
Fix crash where we suggest something prefixed by itself, a range error.

MichaelRFairhurst:
Inside `<div *foo=" ... ">`, we look at the `Foo` directive for inputs
`fooBar`, `fooBaz`, and suggest `bar:`, and `baz:`. However, when `Foo`
has a `foo` input, we may try to suggest it at times. In these times,
it crashes because it gets the suffix (empty string) and attempts to
decapitalize the first character (there is no first character), and it
crashes.

We have no business suggesting it. That is to say, we *could* handle
the crash and suggest ':', but that would be a bogus suggestion anyway.

We already do a prefix check, simply a matter of adding an equality
check too.

### DIFF
--- a/angular_analyzer_plugin/lib/src/completion.dart
+++ b/angular_analyzer_plugin/lib/src/completion.dart
@@ -859,10 +859,17 @@ class TemplateCompleter {
         if (usedInputs.contains(input)) {
           continue;
         }
+
         // edge case. Don't think this comes up in standard.
         if (!input.name.startsWith(templateAttr.prefix)) {
           continue;
         }
+
+        // Unlick ngForTrackBy becoming trackBy, ngFor can't become anything.
+        if (input.name == templateAttr.prefix) {
+          continue;
+        }
+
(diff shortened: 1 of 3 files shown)


angular_analyzer_plugin/lib/src/completion.dart
```diff
@@ -859,10 +859,17 @@ class TemplateCompleter {
         if (usedInputs.contains(input)) {
           continue;
         }
+
         // edge case. Don't think this comes up in standard.
         if (!input.name.startsWith(templateAttr.prefix)) {
           continue;
         }
+
+        // Unlick ngForTrackBy becoming trackBy, ngFor can't become anything.
```
bwilkerson: "Unlick" --> "Unlink"?

MichaelRFairhurst: ah, "unlike" even! "unlick" definitely not correct here...!

PR #476, opened 2017-12-08
Sync master into SDK_AT_HEAD

MichaelRFairhurst:
will be rolling SDK_AT_HEAD into google3 soon, want to make sure SDK_AT_HEAD is up to date with master.


MichaelRFairhurst:
Very strange that googlebot doesn't like this...

MichaelRFairhurst:
trying:

I signed it!

But I can actually merge this either way as an owner, it seems.

PR #482, opened 2017-12-22
Support (focusin) and (focusout), taking care of its nonexist getter

MichaelRFairhurst:
It's not a part of dart:html, but it works inside angular by binding to
events by name. Find its event type (FocusEvent) manually, after
performing the usual standard html resolution.

Since it doesn't have a getter, it doesn't have a range. Test that
navigation won't crash.

### DIFF
--- a/angular_analyzer_plugin/lib/src/standard_components.dart
+++ b/angular_analyzer_plugin/lib/src/standard_components.dart
@@ -169,10 +169,27 @@ class BuildStandardHtmlComponentsVisitor extends RecursiveAstVisitor {
     'tabIndex': 'tabindex',
   };
 
+  static const missingOutputs = const {
+    'focusin': 'FocusEvent',
+    'focusout': 'FocusEvent',
+  };
+
   ClassElement classElement;
 
   BuildStandardHtmlComponentsVisitor(this.components, this.events,
       this.attributes, this.source, this.securitySchema);
+
+  @override
+  void visitCompilationUnit(ast.CompilationUnit unit) {
+    super.visitCompilationUnit(unit);
+
+    missingOutputs.forEach((name, type) {
+      final namespace = unit.element.library.publicNamespace;
+      final ClassElement eventClass = namespace.get(type);
+      events[name] = new OutputElement(
+          name, null, null, unit.element.source, null, null, eventClass.type);
+    });
+  }
 
   @override
   void visitClassDeclaration(ast.ClassDeclaration node) {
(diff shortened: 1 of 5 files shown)


angular_analyzer_plugin/test/angular_driver_test.dart
```diff
@@ -245,6 +245,20 @@ class BuildStandardHtmlComponentsTest extends AbstractAngularTest {
       final outputElement = outputElements['dden'];
       expect(outputElement, isNull);
     }
+    {
+      // missing from dart:html, and supplied manually (with no getter)
+      final outputElement = outputElements['focusin'];
+      expect(outputElement, isNotNull);
+      expect(outputElement.eventType, isNotNull);
+      expect(outputElement.eventType.toString(), equals('FocusEvent'));
```
pq: For brevity, I think you can elide the `equals`:


    expect(outputElement.eventType.toString(), 'FocusEvent');

angular_analyzer_plugin/test/angular_driver_test.dart
```diff
@@ -245,6 +245,20 @@ class BuildStandardHtmlComponentsTest extends AbstractAngularTest {
...
+    }
+    {
+      // missing from dart:html, and supplied manually (with no getter)
+      final outputElement = outputElements['focusout'];
+      expect(outputElement, isNotNull);
+      expect(outputElement.eventType, isNotNull);
+      expect(outputElement.eventType.toString(), equals('FocusEvent'));
```
pq: (and here)

angular_analyzer_plugin/lib/src/standard_components.dart
```diff
@@ -169,11 +169,28 @@ class BuildStandardHtmlComponentsVisitor extends RecursiveAstVisitor {
...
 
   BuildStandardHtmlComponentsVisitor(this.components, this.events,
       this.attributes, this.source, this.securitySchema);
 
+  @override
+  void visitCompilationUnit(ast.CompilationUnit unit) {
+    super.visitCompilationUnit(unit);
+
+    missingOutputs.forEach((name, type) {
+      final namespace = unit.element.library.publicNamespace;
```
pq: Naive question: is there any chance the library will be `null` here?  (Or for that matter, `element`?)

MichaelRFairhurst: Should be provided by the SDK, so in theory it could be but that would mean I screwed up my SDK constraint.

angular_analyzer_plugin/test/angular_driver_test.dart
MichaelRFairhurst: Thanks!

PR #485, opened 2018-01-03
Allow custom events with custom types to be specified.

MichaelRFairhurst:
```
analyzer:
  plugins:
    angular:
      enabled: true
      custom_events:
        doodle:
          type: DoodleEvent
          path: 'package:doodle/events.dart'
        poodle:
          type: PoodleEvent
          path: 'package:doodle/events.dart'
```

These don't yet correlate to tags, that could be next up, either
allowing a tag name in the custom_events section, or allowing
custom_events within a tag section, or both.

Still haven't fully decoupled [Options] from [FileTracker], but at
least the new hashes are *less* coupled, and I have some ideas to go
the rest of the way cleanly.

Gather the custom events inside the driver, and make it look like a
part of standard html. Track the source of the options file and the
name offsets of the custom events so that it they be supplied during
completion etc.

Currently doesn't re-resolve the custom event type. We'll see how many
people need it to react to change. I have a feeling that won't be
important, since this is mostly used for polymer users, where polymer
won't be under active development any more.

Also no errors reported yet for misconfigured analysis options. That
has to be a new pipeline unfortunately.

### DIFF
--- a/angular_analyzer_plugin/lib/src/angular_driver.dart
+++ b/angular_analyzer_plugin/lib/src/angular_driver.dart
@@ -5,7 +5,6 @@ import 'package:analyzer/src/dart/analysis/byte_store.dart';
 import 'package:analyzer/error/listener.dart';
 import 'package:analyzer/src/generated/sdk.dart';
 import 'package:analyzer/src/generated/source.dart';
-import 'package:analyzer/src/generated/resolver.dart' show TypeProvider;
 import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/type.dart';
 import 'package:analyzer/error/error.dart';
@@ -279,7 +278,7 @@ class AngularDriver
   }
 
   Future<OutputElement> getCustomOutputElement(
-      CustomEvent event, TypeProvider typeProvider) async {
+      CustomEvent event, DartType dynamicType) async {
     OutputElement defaultOutput() => new OutputElement(
         event.name,
         event.nameOffset,
@@ -287,7 +286,7 @@ class AngularDriver
         options.source,
         null,
         null,
-        typeProvider.dynamicType);
+        dynamicType);
 
     final typePath =
         event.typePath ?? (event.typeName != null ? 'dart:core' : null);
@@ -310,9 +309,8 @@ class AngularDriver
     if (typeElement is ClassElement) {
       var type = typeElement.type;
       if (type is ParameterizedType) {
-        type = type.instantiate(type.typeParameters
-            .map((p) => p.bound ?? typeProvider.objectType)
-            .toList());
+        type = type.instantiate(
+            type.typeParameters.map((p) => p.bound ?? dynamicType).toList());
       }
       return new OutputElement(event.name, event.nameOffset, event.name.length,
           options.source, null, null, type);
@@ -340,8 +338,8 @@ class AngularDriver
           components, standardEvents, attributes, source, securitySchema));
 
       for (final event in options.customEvents.values) {
-        customEvents[event.name] =
-            await getCustomOutputElement(event, result.typeProvider);
+        customEvents[event.name] = await getCustomOutputElement(
+            event, result.typeProvider.dynamicType);
       }
 
       standardHtml = new StandardHtml(
--- a/angular_analyzer_plugin/lib/src/angular_driver.dart
+++ b/angular_analyzer_plugin/lib/src/angular_driver.dart
@@ -5,7 +5,9 @@ import 'package:analyzer/src/dart/analysis/byte_store.dart';
 import 'package:analyzer/error/listener.dart';
 import 'package:analyzer/src/generated/sdk.dart';
 import 'package:analyzer/src/generated/source.dart';
+import 'package:analyzer/src/generated/resolver.dart' show TypeProvider;
 import 'package:analyzer/dart/element/element.dart';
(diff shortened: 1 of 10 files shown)


angular_analyzer_plugin/lib/src/angular_driver.dart
```diff
@@ -276,6 +278,53 @@ class AngularDriver
...
+      return defaultOutput();
+    }
+
+    final typeElement =
+        typeResult.libraryElement.publicNamespace.get(event.typeName);
+    if (typeElement is ClassElement) {
+      var type = typeElement.type;
+      if (type is ParameterizedType) {
+        type = type.instantiate(type.typeParameters
+            .map((p) => p.bound ?? typeProvider.objectType)
```
leonsenft: Isn't an unbound generic considered `dynamic`?

angular_analyzer_plugin/lib/src/options.dart
```diff
@@ -1,12 +1,60 @@
...
+import 'package:analyzer/src/generated/source.dart';
 
 class AngularOptions {
-  AngularOptions({this.customTagNames});
-  factory AngularOptions.from(String contents) =>
-      new _OptionsBuilder(contents).build();
+  AngularOptions({this.customTagNames, this.customEvents, this.source});
+  factory AngularOptions.from(Source source) =>
+      new _OptionsBuilder(null, source).build();
+  factory AngularOptions.fromString(String content, Source source) =>
```
leonsenft: What is the difference between `content` and `source`? Perhaps document these?

angular_analyzer_plugin/test/angular_driver_test.dart
```diff
@@ -299,6 +301,348 @@ class BuildStandardHtmlComponentsTest extends AbstractAngularTest {
...
+      expect(event.getter, isNull);
+      expect(event.eventType, isNotNull);
+      expect(event.eventType.toString(), 'bool');
+      expect(
+          event.eventType.element.source.fullName, '/sdk/lib/core/core.dart');
+    }
+  }
+
+  // ignore: non_constant_identifier_names
+  Future test_customEvents_noSource_dynamic() async {
```
leonsenft: So if a custom event's source doesn't exist (whether the file or definition is missing), do we still allow it to be used as dynamic without warning?

angular_analyzer_plugin/lib/src/angular_driver.dart
MichaelRFairhurst: Hmm. I had thought it was Object in strong mode, so that the final method call in this example is flagged as not allowed (if it were dynamic, all bets would be off).

```
  x = new List();
  x.add(1);
  x[0].supercallifragalisticexpialidocious();
```

but trying this locally, it looks like it succeeds in strong mode....Maybe dart 2 it will fail?

I guess a followup question is, if someone refers to say, `DataEvent<T>`, is it best if they can do `$event.data.supercallifragalisticexpialidocious` or not?

Probably better if they can, at least for now, since they can't cast inside templates.

WDUT?

angular_analyzer_plugin/lib/src/options.dart
MichaelRFairhurst: Ah yes, in fact, I can do one better by adding `@visibleForTesting`!

angular_analyzer_plugin/test/angular_driver_test.dart
MichaelRFairhurst: Unfortunately for now I don't have the infrastructure plumbed up to report erroneous configs. That will have to come later...for now it will fail silently. :(

angular_analyzer_plugin/lib/src/angular_driver.dart
leonsenft: Yes, it's definitely treated as dynamic in strong mode currently. I'm not aware of whether this is changing or not in Dart 2. I think it's best to remain consistent with current behavior, which is treating an unspecified type argument as dynamic; there's plenty of code that behaves this way in Angular already. The inability to cast inside templates is a good motivation for making it dynamic as well.

angular_analyzer_plugin/lib/src/options.dart
leonsenft: :+1:

angular_analyzer_plugin/test/angular_driver_test.dart
leonsenft: Fail silently as in treat an event with that name as an error, or treat it as dynamic?

MichaelRFairhurst: Ah, yes, sorry, the latter.

MichaelRFairhurst:
@leonsenft Changes in, does this LGTY?

leonsenft:
@MichaelRFairhurst What about the default type of an unbound type parameter? Shouldn't it be changed to `dynamic` to match the semantics of the language? Or do you think being more restrictive encourages people to use it properly with a bound generic?

angular_analyzer_plugin/lib/src/angular_driver.dart
MichaelRFairhurst: Done!

MichaelRFairhurst:
@leonsenft Yes! I fixed that too, the dynamic type parameter. Though it looks like it was missing from my original commit...did a commit --amend to include!

PR #487, opened 2018-01-19
Publish 0.0.12 with latest analyzer & angular_ast

MichaelRFairhurst:


kevmoo:
@MichaelRFairhurst – Update this guy when you're done? https://github.com/dart-lang/angular/blob/master/angular/tools/analyzer_plugin/pubspec.yaml

MichaelRFairhurst:
@kevmoo will do, however, I'm not sure how critical it is since I think pub will pull in this new version automatically just by it being published?

kevmoo:
> pub will pull in this new version automatically just by it being published?

Yup. Sorry. I think `^0.0.10` ==> `>=0.0.10 <0.1.0`

PR #488, opened 2018-01-23
Test that strong mode works, and use `TypeSystem` so `FutureOr` does too

MichaelRFairhurst:


angular_analyzer_plugin/test/mock_sdk_test.dart
```diff
@@ -0,0 +1,26 @@
...
+@reflectiveTest
+class MockSdkTest extends AbstractAngularTest {
+  // ignore: non_constant_identifier_names
+  Future test_futureOr() async {
+    final dartResult =
+        await dartDriver.getResult('$sdkRoot/lib/async/async.dart');
+    expect(dartResult.errors, isEmpty);
+    expect(
+        dartResult.libraryElement.exportNamespace.get('FutureOr'), isNotNull);
+    print(dartResult.libraryElement.context.typeProvider.futureOrType);
```
scheglov: D you really want to have this test with debug output?

MichaelRFairhurst: Definitely not! Good catch, thanks :)

PR #515, opened 2018-03-12
Correctness/Performance fix: Use FileContentOverlay in navigation

MichaelRFairhurst:
This is a big performance regression in IntelliJ when inside HTML (where
content info isn't as nicely cached for us by the driver). Before
serving completions, for instance, the navigation would be reading in
large files and computing their line infos which was like 90%+ of CPU
time.

By sharing the overlay, navigation is now 0.16% of CPU time and
completions are much snappier.

### DIFF
--- a/angular_analyzer_plugin/lib/src/navigation.dart
+++ b/angular_analyzer_plugin/lib/src/navigation.dart
@@ -6,8 +6,13 @@ import 'package:analyzer/dart/element/element.dart' as engine;
 import 'package:analyzer/src/generated/source.dart';
 import 'package:angular_analyzer_plugin/src/model.dart';
 import 'package:angular_analyzer_plugin/src/navigation_request.dart';
+import 'package:analyzer/src/dart/analysis/file_state.dart';
 
 class AngularNavigation implements NavigationContributor {
+  final FileContentOverlay _contentOverlay;
+
+  AngularNavigation(this._contentOverlay);
+
   @override
   void computeNavigation(
       NavigationRequest baseRequest, NavigationCollector collector,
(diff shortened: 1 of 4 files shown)


angular_analyzer_plugin/lib/src/navigation.dart
```diff
@@ -93,7 +98,8 @@ class AngularNavigation implements NavigationContributor {
       }
 
       final lineInfo = element.compilationElement?.lineInfo ??
-          new LineInfo.fromContent(element.source.contents.data);
+          new LineInfo.fromContent(
```
scheglov: Should we cache LineInfo for sources in this method?

Or even better, is there something like FileState in Angular plugin?

angular_analyzer_plugin/lib/src/navigation.dart
```diff
@@ -93,7 +98,8 @@ class AngularNavigation implements NavigationContributor {
...
+              _contentOverlay[element.source.fullName] ?? "");
```
scheglov: Also it might be a question of correctness too.
`element.source.contents` reads file content, so if there was overlay, the file content is different.

MichaelRFairhurst: Absolutely!

angular_analyzer_plugin/lib/src/navigation.dart
MichaelRFairhurst: At the moment it's not cached and not a problem.

We do have a file tracker, it mostly just tracks the equivalent of part/lib relationships. I could see about having it track this as well...

PR #516, opened 2018-03-12
Begin changelog, tracked changes back to 0.0.10

MichaelRFairhurst:


CHANGELOG.md
```diff
@@ -0,0 +1,103 @@
+# CHANGELOG
```
devoncarew: I think we can omit the CHANGELOG title

CHANGELOG.md
```diff
@@ -0,0 +1,103 @@
...
+          - baz
+```
+
+Most errors related to custom tags are suppressed, because custom tags are often
+handled by other frameworks (ie, polymer).
+
+# 0.0.10
+
+Started changelog.
+
```
devoncarew: nit: trailing newlines

CHANGELOG.md
MichaelRFairhurst: Done

CHANGELOG.md
MichaelRFairhurst: Done

PR #521, opened 2018-03-21
Make AngularDriver less prone to race conditions with smaller API.

MichaelRFairhurst:
Public members resolveHtml, resolveDart have to be atomic to be
guaranteed to work well and at the very least easy to read, reason
about, and debug. In addition, resolveHtml has an order-of-operations
requirement that it comes after all dart has been fully analyzed.

Change the requestErrors APIs (which we only used in the old API) to
request full [DirectiveResult]s, and use that everywhere that we used to
use resolve(Dart|Html), because that was made free of race conditions.
One caveat here is that everywhere resolve(Dart|Html) was used, it
required cache busting. This means that the deprecated [requestErrors]
APIs will be slower as those didn't used to, and don't need to, cache
bust. Since they're deprecated this isn't a problem.

Another note, is that we can't share a scheduler with the lower-level
driver if we want [performWork] to be atomic (which we do). That's
because we request data from the underlying driver which won't be
complete until that driver can [performWork], which will never happen if
the scheduler is waiting for the angular driver to finish. Therefore,
rather than giving up on having [performWork] being atomic inside
[AngularDriver], create a new [AnalysisDriverScheduler] for the dart
driver so that we don't have this problem. This does mean we have two
"competing" schedulers, but I think that's fine for now anyway.

This was causing issues on windows because order-of-operations there was
different and so it was behaving in weird ways in response to workflows
that were previously unexercized. Should be fixed now, and more reliable
in response to *all* different workflows.

### DIFF
--- a/angular_analyzer_plugin/lib/src/angular_driver.dart
+++ b/angular_analyzer_plugin/lib/src/angular_driver.dart
@@ -49,8 +49,8 @@ class AngularDriver
   final _addedFiles = new LinkedHashSet<String>();
   final _dartFiles = new LinkedHashSet<String>();
   final _changedFiles = new LinkedHashSet<String>();
-  final _requestedDartFiles = <String, List<Completer>>{};
-  final _requestedHtmlFiles = <String, List<Completer>>{};
+  final _requestedDartFiles = <String, List<Completer<DirectivesResult>>>{};
+  final _requestedHtmlFiles = <String, List<Completer<DirectivesResult>>>{};
   final _filesToAnalyze = new HashSet<String>();
   final _htmlFilesToAnalyze = new HashSet<String>();
   final ByteStore byteStore;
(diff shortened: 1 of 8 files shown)


MichaelRFairhurst:
Confirmed this works on windows well

angular_analyzer_plugin/lib/src/angular_driver.dart
```diff
@@ -214,63 +232,53 @@ class AngularDriver
...
-        completers.forEach((completer) =>
-            completer.complete(result?.errors ?? <AnalysisError>[]));
-      }, onError: (e) {
+        // After whichever resolution is complete, push errors.
+        completers.forEach((completer) => completer.complete(result));
+      } catch (e, st) {
         completers.forEach((completer) => completer.completeError(e));
-      });
+      }
+      ;
```
scheglov: Remove?

MichaelRFairhurst: Done

PR #535, opened 2018-03-30
Lexically and unintelligently accept `===` by replacing it with `== `.

MichaelRFairhurst:
This is not the cleanest solution from a "language theory" perspective.
I did try extending [parseEqualityExpression] to accept the new
operator, but it isn't particularly clean since equality operators have
a few micro details (allowing super==, disallowing x == y == z). Also,
since === has the same precedence as ==, they must share a loop which
means copying the lower implementation, which seemed like too much
coupling, on top of the fact that its already extending something that
will have a limited lifetime.

By replacing === with ==, we may see some strange behavior. For
instance, 'x === y === z' will highlight two characters of the second
'===' when it complains about doing equality with an equality.

But the important thing is that it doesn't matter if we butcher string
literals, and there is no other type of token we can be replacing here.
And the semantics of == vs === should be the same statically for users.

### DIFF
--- a/angular_analyzer_plugin/test/resolver_test.dart
+++ b/angular_analyzer_plugin/test/resolver_test.dart
@@ -5319,7 +5319,7 @@ class FutureOrApis {
   }
 
   // ignore: non_constant_identifier_names
-  Future solo_test_tripleEq() async {
+  Future test_tripleEq() async {
     _addDartSource(r'''
 import 'dart:async';
 @Component(selector: 'a', templateUrl: 'test_panel.html')
--- a/angular_analyzer_plugin/test/resolver_test.dart
+++ b/angular_analyzer_plugin/test/resolver_test.dart
@@ -5318,6 +5318,22 @@ class FutureOrApis {
     errorListener.assertNoErrors();
   }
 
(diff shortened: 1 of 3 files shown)


angular_analyzer_plugin/test/resolver_test.dart
```diff
@@ -5318,6 +5318,22 @@ class FutureOrApis {
     errorListener.assertNoErrors();
   }
 
+  // ignore: non_constant_identifier_names
+  Future solo_test_tripleEq() async {
```
bwilkerson: You probably intended to remove the `solo_` prefix.

MichaelRFairhurst: Thanks! Yep!

MichaelRFairhurst:
This syntax is also easily droppable, the main thing is that until we know whether we are going to make our own parser or not, we don't want to force users to stop using it.

Yep its pretty ugly, and its worse for the compiler team because where pipes could be handled with a simple tokenizer, === has a precedence right smack dab in the middle of everything else and would therefore be much harder to handle that way.

PR #547, opened 2018-04-11
Add support for `angular:` toplevel in analysis_options.yaml.

MichaelRFairhurst:
Also bring back the merge code that was there before for when I handled
the case of configuring `plugins: angular` as well as
`plugins:angular_analyzer_plugin`. That code didn't make sense when that
would result in two plugins loaded at once, however, now it is
reasonable (albeit deprecated) to have both the angular top level, and
the `options: angular(analyzer_plugin)?` section configured at the same
time.

Its not critical we handle it perfectly, but it is important that we
handle it consistently until that option is removed.


CHANGELOG.md
```diff
@@ -1,13 +1,54 @@
...
+  plugins:
+    - angular
+
+angular:
+  custom_tag_names:
+    - foo
+    - bar
+```
+
+This is encouraged for users on more recent versions than 0.0.14. Support for
```
bwilkerson: Not a problem with the CL, but something to think about. It occurs to me that users don't have any easy way to find out what version of a plugin they're using. I think they can get that info from the status pages, but most users probably don't even know that the status pages exist.

angular_analyzer_plugin/lib/src/options.dart
```diff
@@ -100,37 +101,43 @@ class _OptionsBuilder {
...
+    loadSection('angular') || loadSection('angular_analyzer_plugin');
+
+    resolve();
   }
 
   bool loadSection(String key) {
-    final pluginSection = analysisOptions['analyzer']['plugins'];
+    final pluginsSection = analysisOptions['analyzer']['plugins'];
 
     // Standard case, just turn on the plugin.
```
bwilkerson: This comment is out of date.

angular_analyzer_plugin/lib/src/options.dart
```diff
@@ -100,37 +101,43 @@ class _OptionsBuilder {
...
+    if (pluginsSection is! Map) {
       return false;
     }
 
     // Edge case, a section with a config (such as custom tag names).
-    final specified = pluginSection.containsKey(key);
+    final specified = pluginsSection.containsKey(key);
     if (specified) {
-      angularOptions = pluginSection[key];
+      pluginSection = pluginsSection[key];
```
bwilkerson: I don't understand why we re-assign to a local variable that we never reference again.

Did you intend to remove the `final` from line 115 so that it references the field?

MichaelRFairhurst: Hmm, I knew this would be confusing when I wrote it and it seems I should have listened to that instinct. They're different names...The local variable `pluginsSection` contains all plugins, and the member `pluginSection` contains just the angular config. I'll rename it to, maybe, `allPluginSections`, or something even clearer if I can think of it.

angular_analyzer_plugin/lib/src/options.dart
MichaelRFairhurst: Great catch!

CHANGELOG.md
MichaelRFairhurst: Aha!

I had seen this, but the version was wrong. I thought it was because it was showing the version of angular/the host package.

But no, its just that I have the wrong version in `lib/plugin.dart`. Unfortunately, that's published as `1.0.0-alpha.0`. So I can skip straight to `1.0.0-alpha.1` if I want to reference that.

I'll likely publish this next version as 0.0.15 so that angular's current config of `^0.0.8` works. And then I'll publish the same code as `1.0.0-alpha.1`, and update angular to depend on `1.0.0-alpha.1^` and then at some point push a `1.0.0` stable. Whew.

MichaelRFairhurst:
OK, changed to load the first section it finds and stop there.

PR #557, opened 2018-04-20
Remove (now) unnecessary overrides, from removing AnalysisTarget

MichaelRFairhurst:


angular_analyzer_plugin/lib/src/model.dart
```diff
@@ -726,13 +726,11 @@ class View {
   }
 
   /// The source that contains this view.
-  @override
   Source get source => classElement.source;
```
devoncarew: Are these methods still called?

MichaelRFairhurst: `source` is, but not `librarySource`! Good thinking!

PR #558, opened 2018-04-24
Create more robust synthetic tokens for completion integration.

MichaelRFairhurst:
This won't solve the problem by itself without updating the SDK (PR incoming)...but I'm also curious if this passes on travis, in which case we can commit it to master without issue and support both versions (internal/external) without branching.

When autocompleting a dart snippet like "x" within "{{}}", there is no
parent node and that breaks some logic in the completion engine.
Therefore we add a synthetic parenthesized ast, but previously we had
been giving null begin and end tokens.

Create begin and end tokens via the front-end synthetic classes, and
ensure to link beginToken.next to the user's dart code's beginToken.

### DIFF
--- a/angular_analyzer_plugin/lib/src/completion_request.dart
+++ b/angular_analyzer_plugin/lib/src/completion_request.dart
@@ -9,6 +9,8 @@ import 'package:angular_analyzer_plugin/ast.dart';
 import 'package:angular_analyzer_plugin/src/converter.dart';
 import 'package:angular_analyzer_plugin/src/model.dart';
 import 'package:angular_analyzer_plugin/src/standard_components.dart';
+import 'package:front_end/src/scanner/token.dart'
+    show SyntheticBeginToken, TokenType, SyntheticToken;
 
 class AngularCompletionRequest extends CompletionRequest {
   final List<Template> templates;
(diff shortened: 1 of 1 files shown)


angular_analyzer_plugin/lib/src/completion_request.dart
```diff
@@ -9,6 +9,8 @@ import 'package:angular_analyzer_plugin/ast.dart';
 import 'package:angular_analyzer_plugin/src/converter.dart';
 import 'package:angular_analyzer_plugin/src/model.dart';
 import 'package:angular_analyzer_plugin/src/standard_components.dart';
+import 'package:front_end/src/scanner/token.dart'
+    show SyntheticBeginToken, TokenType, SyntheticToken;
```
stereotype441: Note that the analyzer re-exports TokenType via `package:analyzer/dart/ast/token.dart` so ideally it would be better to import it from there (we are trying to avoid having too many packages depend on front_end while its API is unstable).

Same issue ought to apply to SyntheticBeginToken and SyntheticToken, except that...oops!...the analyzer doesn't re-export them.  I'll make a change to the analyzer that causes it to re-export SyntheticBeginToken and SyntheticToken via `package:analyzer/src/dart/ast/token.dart`, and then once that lands we can change this import statement.

But for now I think it's fine to leave it as is.

MichaelRFairhurst:
Looks like I'll have to close this PR, since it doesn't integrate with the latest version of package:analyzer_plugin. But I'll make a different one that merges into my side branch...

angular_analyzer_plugin/lib/src/completion_request.dart
MichaelRFairhurst: good catch, thanks!

PR #571, opened 2018-05-08
Fix *= selector to work correctly

MichaelRFairhurst:


angular_analyzer_plugin/lib/src/selector.dart
```diff
@@ -156,31 +156,22 @@ class AttributeSelector extends AttributeSelectorBase {
   }
 }
 
-/// The [WildcardSelector] that matches elements that have attributes that start
-/// with the given name, and (optionally) with the given value.
-class WildcardAttributeSelector extends AttributeSelectorBase {
+/// The [AttributeContainsSelector] that matches elements that have attributes
+/// have the given name, and that attribute contains the value of the selector.
```
leonsenft: typo: "elements that have attributes _with_ the given name" ?

angular_analyzer_plugin/lib/src/selector.dart
```diff
@@ -109,6 +109,11 @@ abstract class AttributeSelectorBase extends Selector {
           ? []
           : [nameElement];
 
+  @override
+  bool availableTo(ElementView element) =>
+      !element.attributes.keys.contains(findAttribute(element)) ||
```
leonsenft: I found this condition rather confusing. The `element` passed to `findAttribute` isn't used, but rather the name of the `nameElement` is returned. Perhaps it's the overloaded use of the term 'element' that's confusing me. This isn't an HTML element right, but it's actually the attribute name it's checking for?

Furthermore, wouldn't we want to check that the attributes on the element _do_ contain the selector name?

angular_analyzer_plugin/test/selector_test.dart
```diff
@@ -193,41 +193,66 @@ class AttributeSelectorTest extends _SelectorTest {
...
+
+  // ignore: non_constant_identifier_names
+  void test_noMatch_attributeName() {
+    final selector = new AttributeContainsSelector(nameElement, 'match');
+    when(element.attributes).thenReturn({'unmatched': 'this is matched'});
+    when(element.attributeNameSpans)
+        .thenReturn({'unmatched': _newStringSpan(100, 'unmatched')});
     // verify
     expect(selector.match(element, template), equals(SelectorMatch.NoMatch));
     expect(selector.availableTo(element), true);
```
leonsenft: Actually seeing this case makes me realize I don't understand what `availableTo` is for. I read the doc comment but I didn't still understand it. Mind elaborating please?

angular_analyzer_plugin/test/selector_test.dart
```diff
@@ -696,9 +721,9 @@ class SelectorParserTest {
 
   // ignore: non_constant_identifier_names
   void test_attribute_hasWildcard() {
-    final WildcardAttributeSelector selector =
+    final AttributeContainsSelector selector =
         new SelectorParser(source, 10, '[kind*=pretty]').parse();
-    expect(selector, const isInstanceOf<WildcardAttributeSelector>());
+    expect(selector, const isInstanceOf<AttributeContainsSelector>());
```
leonsenft: Isn't this check redundant if the variable is type annotated with the same type? Won't the previous assignment fail in Dart 2 anyways?

I'm assuming the type annotation is to access the fields declared on that particular type. I've run into this pattern pretty often, and generally end up writing a custom matcher that extends `isInstanceOf`. That might be a bit heavy for this use case. Perhaps type promote after checking `isInstanceOf`, or remove the `isInstanceOf` check altogether and rely on the type system to complain if it isn't the expected type?

angular_analyzer_plugin/test/selector_test.dart
leonsenft: Oh I'm guessing this has something to do with suggestions?

angular_analyzer_plugin/lib/src/selector.dart
MichaelRFairhurst: First: Yep, the `nameElement` is an `Element` in the sense that `ClassElement`s are `Element`s: its a named definition in your code you can navigate to (you can click from `foo` to `[foo]` to see what attributes are matching what directives in your template).

And yeah, 'availableTo' is maybe *too* "intuitive" of a name for the otherwise hard to distill concept.

We want to suggest the attribute when it doesn't appear on the tag, BUT, we also do want to "suggest" the tag when it *does* appear if it matches. That makes it so that `[a=x][b]` will suggest `b` on `<div a=x>` but not on `<div a=y>`. The suggestion phase is a later step, after we know that `[a=x][b]` matches, we find all attribute selectors (`a`, `b`), and we then find only those that aren't already there (`b`).

Perhaps we have everything we need in `match()` to do this -- if we add a new `SelectorMatch.INCOMPATIBLE` then we can maybe distinguish the directives that *could* match from the ones that fundamentally don't, and then maybe wouldn't need `availableTo` at all... That should be a different PR though.

angular_analyzer_plugin/test/selector_test.dart
MichaelRFairhurst: Yep, thanks! Will remove these in a different CL (there are lots of them)

angular_analyzer_plugin/lib/src/selector.dart
leonsenft: Ah I see. So the idea is to only suggest missing selectors, so long as the current set of selectors don't invalidate the entire selector?

MichaelRFairhurst: yep! much better put than what I'd said :)

PR #572, opened 2018-05-08
Better error message for when operators have an operator w/ no value

MichaelRFairhurst:


angular_analyzer_plugin/test/selector_test.dart
```diff
@@ -750,6 +750,42 @@ class SelectorParserTest {
     expect(selector.value, 'bar');
   }
 
+  // ignore: non_constant_identifier_names
+  void test_attribute_regularOperator_noValue() {
+    try {
+      new SelectorParser(source, 0, '[foo=]').parse();
+    } on FormatException catch (e) {
+      expect(e.message, contains('Unexpected ]'));
```
leonsenft: I think a more actionable error message would be

> Expected a value before ']'

For example `[attr^=]` would yield

```
Error: Expected value before ']'
    [attr^=]
    ^^^^^^^^
```

Or maybe

> Expected a value between '^=' and ']'

What do you think?

PR #583, opened 2018-06-24
Publish 0.0.16

MichaelRFairhurst:


devoncarew:
It looks like travis is failing here: https://travis-ci.org/dart-lang/angular_analyzer_plugin/jobs/396171501#L1492

Since the dart --preview-dart-2 default changed, I think here: https://github.com/dart-lang/angular_analyzer_plugin/blob/master/tools/travis.sh#L43 you can pass in `--no-preview-dart-2 --checked`, and for the next line, I'd remove the preview-dart-2 flag, and pass in `--enable-asserts`. You could also choose to stop testing in dart 1 mode...

MichaelRFairhurst:
I opened #585 and #584 and assigned to kevin to review...the two together should fix the build and I agree I'd like that to happen before publishing this

Didn't include `--enable-asserts`, though, that is a good catch!

PR #585, opened 2018-06-24
Fix travis build: don't specify --checked without --no-preview-dart-2

MichaelRFairhurst:


tools/travis.sh
```diff
@@ -40,6 +40,6 @@ fi
 dartanalyzer lib test
 
 # Run the actual tests
-dart --checked test/test_all.dart # in dart 1 mode
+dart --no-preview-dart-2 --checked test/test_all.dart # in dart 1 mode
 dart --preview-dart-2 test/test_all.dart # and dart 2 mode
```
devoncarew: Unfortunately, the `--preview-dart-2` flag does't exist in the latest vm...

MichaelRFairhurst: ah, thanks. That seems premeditative to me!

devoncarew: It would have been much, much better if it had been maintained as a flag while there was still an option to disable the new front end.

MichaelRFairhurst:
Going to close this in favor of one PR which fixes both things at once.

PR #589, opened 2018-07-13
Format with latest dartfmt

MichaelRFairhurst:


MichaelRFairhurst:
@munificent This looks like a bug in the formatter to me, that formatting changed the effective contents of the string.

I'm not picky about the new vs old format, in fact, maybe I should swap these tests out to use an actual templating language like mustache or something. The new approach is probably somewhat cleaner and less magical? And in any case, I want it to be green on travis so I'll roll with what's out! But I figured you'd be interested in the scoop behind this.

munificent:
The first diff (`'ResolvedRange at...`) is because dartfmt now reformats the contents of string interpolation and doesn't allow them to split. The only change is inside the interpolated code, not the string itself.

The second one is a bug (https://github.com/dart-lang/dart_style/issues/711). :( Thanks for the heads up. None of my tests caught this, nor did a run on a big corpus. The good news is that implies it's rare — stuffing a multi-line string inside an interpolation inside another string is pretty weird. The bad news is that it makes an actual change to the user's code.

The really bad part is that dartfmt's sanity checking to prevent that doesn't catch this, because the change is a whitespace change, but to a part of the user's program where whitespace is meaningful. :-/

MichaelRFairhurst:
Yeah no problem! Like I said, I am arguably doing something really suspect that might imply I should be pulling in a templating language to use for these tests. Unfortunately there's no yaml serialization library for dart I can use in this case either.

Figured you'd want to see!

MichaelRFairhurst:
ping @scheglov or @munificent for approval

PR #592, opened 2018-07-20
Fix constants in generated files

MichaelRFairhurst:


MichaelRFairhurst:
Broken for unrelated reasons. Landing anyway

PR #593, opened 2018-07-20
Upgrade package analyzer

MichaelRFairhurst:


MichaelRFairhurst:
Yep, these are all the commits from SDK_AT_HEAD which kept the plugin working against the latest analyzer, so those need to be published along with the pubspec update.

This makes review pretty crappy though. Hmm. I'll instead add the new commit into SDK_AT_HEAD, and then I can merge SDK_AT_HEAD into master. That will be easier to review.

PR #596, opened 2018-07-23
upgrade deps and sdk dep; resolve issues

devoncarew:
- upgrade deps on `package:analyzer` and `package:analyzer_plugin`
- widen the sdk dep to include `2.0.0`
- fix some issues due to rolling to the latest package:analyzer
- prep for publishing (so that we have a published package which doesn't use upper case sdk constants)

@MichaelRFairhurst, I don't think this follows the branch strategy that you have for this repo; happy to take advice here about how to better submit this, or have the PR chopped up into other PRs as necessary to support a new publish

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ angular2-dart-analyzer.iml
 
 .packages
 .pub
+.dart_tool
 packages
 pubspec.lock
 deps/sdk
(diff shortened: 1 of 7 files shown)


devoncarew:
closing as most of this has already landed

PR #599, opened 2018-07-25
address a few analysis warnings

devoncarew:
- address a few analysis warnings

This reduces the analysis issues reported from ~130 to 32.

Also, I notice that this package depends on package:tuple. That package has not updated its version constraint to include dart 2. We may want to in-line our use of the package (the Tuple2 and Tuple4 classes) so the angular plugin doesn't break once pub is updated to be strict about sdk deps.

@MichaelRFairhurst

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ angular2-dart-analyzer.iml
 
 .packages
 .pub
+.dart_tool/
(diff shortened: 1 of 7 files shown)


devoncarew:
Thanks!

PR #601, opened 2018-07-27
Remove dependency on package:tuple by infilling the used behavior

MichaelRFairhurst:


angular_analyzer_plugin/lib/src/converter.dart
```diff
@@ -15,7 +15,9 @@ import 'package:angular_analyzer_plugin/src/ignoring_error_listener.dart';
 import 'package:angular_analyzer_plugin/src/strings.dart';
 import 'package:angular_analyzer_plugin/errors.dart';
 import 'package:meta/meta.dart';
-import 'package:tuple/tuple.dart';
+// Does not yet support Dart 2
```
devoncarew: I generally try not to check in commented out code. Perhaps track moving back to `package:tuple` in an issue instead?

angular_analyzer_plugin/lib/src/tuple.dart
```diff
@@ -0,0 +1,16 @@
+// package:tuple does not yet support Dart 2. In the meantime, hardcode the two
```
devoncarew: nit: missing a copyright header

angular_analyzer_plugin/lib/src/converter.dart
MichaelRFairhurst: Hmm, I do this often with TODOs, as in the following:

// TODO: blah blah
// old_code()
new_code();

But I'm open to your reasoning about commented out code. Is it that it gets stale, or adds clutter, or just that it looks like debugging rather than production changes? Maybe a bit of all of the above :)

Can do.

angular_analyzer_plugin/lib/src/tuple.dart
MichaelRFairhurst: I don't yet have any copyright headers!

I'll add them in another CL. I also want to sort members on all my files. Can easily create both CLs in one go I think.

angular_analyzer_plugin/lib/src/converter.dart
devoncarew: Yeah, it's a policy in the flutter repo, I think coming from chromium conventions. I check commented out code in less often now, but there's always exceptions to a rule.

Issue #602, opened 2018-08-03
Not working?


brianegan:
Hey hey :)

I'd love to use this tool, but for some reason it's not quite working for me. I'm not sure what information would be helpful to reproduce, any advice on logs I could provide, etc would be appreciated.

### Repro steps:

  1. Open Webstorm 2018.2
  2. Create a new AngularDart project from the Stagehand template
  3. Open up `analysis_options.yaml` file and enable the plugin.
  4. Restart the editor (instructions found here: https://medium.com/dartlang/announcement-ide-tooling-for-angulardart-5-de01ecccec9d)
  5. Wait a few seconds for the analyzer to download install the plugin (mentioned in the blog post)
  6. Run pub get again, just in case

### Version info:

  - Dart 2.0.0-dev.69.5
  - angular_analyzer_plugin-0.0.17
  - angular-5.0.0-beta+3

### Things I tried

  - Removing everything from the `.dartServer` folder (mentioned in a comment on the blog post)
  - Tried a few different `analysis_options.yaml` configurations

### Screenshot

<img width="1072" alt="screen shot 2018-08-03 at 11 46 26 am" src="https://user-images.githubusercontent.com/126604/43636722-32c3f8f4-9713-11e8-8767-d27a66211f7d.png">

devoncarew:
In order to troubleshoot, can you:
- delete `~/.dartServer/.plugin_manager/` (we store a cached version of the angular analysis plugin there, and it's `pub get` dependencies)
- re-run webstorm; from the gear menu of the dart analysis view, choose 'View analyzer diagnostics'. Then, go to the 'Plugins' page. You should see the angular analysis plugin in use; if there are any issues, you'll likely see an exception on that page.

brianegan:
Thanks! Ok, just completed both steps, and unfortunately it didn't solve the problem, nor does it seem to indicate any errors.

Here is the info from the status page:

```
Status
General status and diagnostics for the analysis server.

Status
Preview-dart-2: true
Use fasta parser: false
Use common front end: false
Instrumentation enabled: false
Server process ID: 40181
Versions
Analysis server version: 1.20.3
Dart SDK: 2.0.0-dev.69.5 (Tue Jul 31 15:05:14 2018 +0200) on "macos_x64"
```

And here is the info from the plugins page:

```
Plugins in use.

Analysis plugins
angular-5.0.0-beta+3
[...]

PR #605, opened 2018-08-09
Fix issue with dynamic pipe argument.

MichaelRFairhurst:
I think this is behind #602


angular_analyzer_plugin/test/angular_driver_test.dart
```diff
@@ -1244,6 +1244,25 @@ class PipeA extends PipeTransform{
     errorListener.assertNoErrors();
   }
 
+  // ignore: non_constant_identifier_names
+  @soloTest
```
bwilkerson: You probably want to remove this.

PR #606, opened 2018-08-10
Sort members on all files

MichaelRFairhurst:


devoncarew:
Looks like you have a fix for the build failures in https://github.com/dart-lang/angular_analyzer_plugin/pull/607? If so, I'd land that and merge master into this PR, so it can go green before landing.

PR #621, opened 2018-08-20
First pass at pipe support, check return type of pipes.

MichaelRFairhurst:
Some issues with pipe parsing offsets, and deserialization fixed.

Otherwise, we stick the pipe name & args into the AST properties, and
our already existing override of ResolverVisitor can find the synthetic
nodes for pipes and set the propagated type accordingly.

Haven't yet handled arguments -- I likely will run into issues where the
argument expressions need to be hit by each of the visitors. So I may be
better off doing a synthetic function call node that's treated
specially, rather than the previous cast technique which was more for
making an AST node that would result in the right analysis. That's no
longer necessary if we're hijacking analysis, and we want the children
of the pipe (when the pipe has arguments) to be typechecked safely which
means its probably best to make them first-class citizens.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Unpublished
 
+- Typecheck the results of pipe expressions and the existence of a matching
+  pipe. Arguments are not yet typechecked.
+
 ## 0.0.17+3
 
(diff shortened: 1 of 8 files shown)


MichaelRFairhurst:
I realized that typechecking the input is easy, its only typechecking the optional arguments that's hard.

PR #623, opened 2018-08-22
Merge master into SDK_AT_HEAD

MichaelRFairhurst:
this just merges master into SDK_AT_HEAD


angular_analyzer_plugin/lib/src/ng_expr_parser.dart
```diff
@@ -25,21 +26,23 @@ class NgExprParser extends Parser {
...
       beforePipeToken ??= _currentToken.previous;
       getAndAdvance();
-      parsePipeExpression();
-    }
-    if (beforePipeToken != null) {
-      final asToken = new KeywordToken(Keyword.AS, 0);
-      final dynamicIdToken =
-          new StringToken(TokenType.IDENTIFIER, "dynamic", 0);
+      final pipeData = parsePipeExpression();
+      if (beforePipeToken != null) {
```
stereotype441: Is this "if" redundant now, because we are inside the while loop, thus line 27 `beforePipeToken ??= _currentToken.previous` is guaranteed to have already executed?

angular_analyzer_plugin/lib/src/ng_expr_parser.dart
```diff
@@ -48,11 +51,15 @@ class NgExprParser extends Parser {
   /// Return the resolved left-hand expression as a dynamic type.
   ///
   ///     pipeExpression ::= identifier[':' expression]*
-  void parsePipeExpression() {
-    parseIdentifierList();
+  Tuple2<SimpleIdentifier, List<Expression>> parsePipeExpression() {
```
stereotype441: IMHO tuples should only be used in situations where there is a strong convention for regarding the underlying the data as a pair (e.g. key/value pairs when iterating through a dictionary).  For this case I think it would be much more readable to declare a fresh class so that references to the identifier and expression list go through properties that have comprehensible names, e.g.:

```dart
class PipeData {
  final SimpleIdentifier name;
  final List<Expression> args;
  PipeData(this.name, this.args);
}
```

That should make the code at lines 43-44 easier to understand.

angular_analyzer_plugin/lib/src/ng_expr_parser.dart
```diff
@@ -48,11 +51,15 @@ class NgExprParser extends Parser {
...
+    final identifiers = parseIdentifierList();
+    final expressions = <Expression>[];
     while (_currentToken.type == TokenType.COLON) {
       getAndAdvance();
-      parseExpression2();
+      expressions.add(parseExpression2());
     }
+
+    return Tuple2<SimpleIdentifier, List<Expression>>(
+        identifiers.first, expressions);
```
stereotype441: What happens if `identifiers` contains more than one element?  Is that allowed?  If it's allowed, it seems odd that we discard all identifiers except the first; if it's not, it seems odd that we don't report an error.

angular_analyzer_plugin/lib/src/resolver.dart
```diff
@@ -107,17 +106,36 @@ class AngularErrorVerifier extends _IntermediateErrorVerifier
...
   void visitAsExpression(AsExpression exp) {
     // This means we generated this in a pipe, and its OK.
     if (exp.asOperator.offset == 0) {
       super.visitAsExpression(exp);
+      final pipeName = exp.getProperty<SimpleIdentifier>('_ng_pipeName');
+      final matchingPipes =
+          pipes.where((pipe) => pipe.pipeName == pipeName.name);
+      if (matchingPipes.length != 1) {
+        errorReporter.reportErrorForNode(
+            AngularWarningCode.PIPE_NOT_FOUND, pipeName, [pipeName]);
```
stereotype441: `PIPE_NOT_FOUND` seems like a good error if `matchingPipes.length` is zero; if it's two or more, shouldn't we report some sort of `AMBIGUOUS_PIPE_NAME` error?

MichaelRFairhurst:
Thanks @stereotype441 for the review!

I can address your codereview comments but this is the wrong place -- this is just to merge master into SDK_AT_HEAD

angular_analyzer_plugin/lib/src/ng_expr_parser.dart
MichaelRFairhurst: I was going to say I disagree, because static typing is enough here.

However, there's no static typing here, so a PipeData class is a good idea.

My general rule is that if there's no good name for it and I'm merely trying to return multiple values, I prefer not to make a bogus class. To me "PipeData" is a pretty lame class name. But I would like having better names than "item1" and "item2" given the lack of typechecking that they're used properly.

MichaelRFairhurst:
OK, @stereotype441 PTAL

This is another routine merge from Master into SDK_AT_HEAD to roll those changes into google3. I don't think it should require heavy review. In fact, I'm considering not using github's PR system to do this at all, or just opening the PR as a formalism but not waiting for review to merge it

PR #628, opened 2018-08-24
Support `<ng-container>` properly.

MichaelRFairhurst:
Also refactored `converter.dart` to share opening/close (name)span logic
across the various special tag types (`<normal>`, `<template`,
`<ng-content>`, `<ng-container>`.

Add an assert that all `AstNode`s get matched so this doesn't
transparently happen again as new AngularAst nodes get added.

### DIFF
--- a/angular_analyzer_plugin/lib/src/converter.dart
+++ b/angular_analyzer_plugin/lib/src/converter.dart
@@ -397,78 +397,41 @@ class HtmlTreeConverter {
         references: node.references,
         stars: node.stars,
       )..sort((a, b) => a.offset.compareTo(b.offset));
-      final closeComponent = node.closeComplement;
-      SourceRange openingSpan;
       SourceRange openingNameSpan;
-      SourceRange closingSpan;
-      SourceRange closingNameSpan;
 
-      if (node.isSynthetic) {
-        openingSpan = _toSourceRange(closeComponent.beginToken.offset, 0);
-        openingNameSpan = openingSpan;
-      } else {
-        openingSpan = _toSourceRange(
-            node.beginToken.offset, node.endToken.end - node.beginToken.offset);
+      if (!node.isSynthetic) {
         openingNameSpan = new SourceRange(
             (node as ParsedElementAst).identifierToken.offset,
             (node as ParsedElementAst).identifierToken.lexeme.length);
       }
-      // Check for void element cases (has closing complement)
-      // If closeComponent is synthetic, handle it after child nodes are found.
-      if (closeComponent != null && !closeComponent.isSynthetic) {
-        closingSpan = _toSourceRange(closeComponent.beginToken.offset,
-            closeComponent.endToken.end - closeComponent.beginToken.offset);
-        closingNameSpan =
-            new SourceRange(closingSpan.offset + '</'.length, localName.length);
-      }
 
-      final element = new ElementInfo(
+      return _elementInfoFromNodeAndCloseComplement(
+        node,
         localName,
-        openingSpan,
-        closingSpan,
-        openingNameSpan,
-        closingNameSpan,
         attributes,
-        findTemplateAttribute(attributes),
+        node.closeComplement,
         parent,
-        isTemplate: false,
+        openingNameSpanOverride: openingNameSpan,
       );
+    } else if (node is ContainerAst) {
+      final attributes = _convertAttributes(
+        stars: node.stars,
+      )..sort((a, b) => a.offset.compareTo(b.offset));
 
-      for (final attribute in attributes) {
-        attribute.parent = element;
-      }
-
-      final children = _convertChildren(node, element);
-      element.childNodes.addAll(children);
-
-      if (!element.isSynthetic &&
-          element.openingSpanIsClosed &&
-          closingSpan != null &&
-          (openingSpan.offset + openingSpan.length) == closingSpan.offset) {
-        element.childNodes.add(new TextInfo(
-            openingSpan.offset + openingSpan.length, '', element, [],
-            synthetic: true));
-      }
-
-      return element;
-    }
-    if (node is EmbeddedContentAst) {
-      final localName = 'ng-content';
+      return _elementInfoFromNodeAndCloseComplement(
+        node,
+        'ng-container',
+        attributes,
+        node.closeComplement,
+        parent,
+      );
+    } else if (node is EmbeddedContentAst) {
       final attributes = <AttributeInfo>[];
-      final closeComplement = node.closeComplement;
-      SourceRange openingSpan;
       SourceRange openingNameSpan;
-      SourceRange closingSpan;
-      SourceRange closingNameSpan;
 
-      if (node.isSynthetic) {
-        openingSpan = _toSourceRange(closeComplement.beginToken.offset, 0);
-        openingNameSpan = openingSpan;
-      } else {
-        openingSpan = _toSourceRange(
-            node.beginToken.offset, node.endToken.end - node.beginToken.offset);
-        openingNameSpan =
-            new SourceRange(openingSpan.offset + '<'.length, localName.length);
+      if (!node.isSynthetic) {
+        openingNameSpan = new SourceRange(
+            node.beginToken.offset + '<'.length, 'ng-content'.length);
         final pnode = node as ParsedEmbeddedContentAst;
         final valueToken = pnode.selectorValueToken;
         if (pnode.selectToken != null) {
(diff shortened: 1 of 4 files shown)


angular_analyzer_plugin/lib/src/resolver.dart
```diff
@@ -370,7 +370,9 @@ class ComponentContentResolver extends AngularAstVisitor {
   bool _isStandardTagName(String name) {
     // ignore: parameter_assignments
     name = name.toLowerCase();
-    return !name.contains('-') || name == 'ng-content';
+    return !name.contains('-') ||
+        name == 'ng-content' ||
+        name == 'ng-container';
```
leonsenft: Why are `ng-content` and `ng-container` considered as HTML tags?

angular_analyzer_plugin/lib/src/converter.dart
```diff
@@ -397,78 +397,41 @@ class HtmlTreeConverter {
...
-        openingNameSpan,
-        closingNameSpan,
         attributes,
-        findTemplateAttribute(attributes),
+        node.closeComplement,
         parent,
-        isTemplate: false,
+        openingNameSpanOverride: openingNameSpan,
       );
+    } else if (node is ContainerAst) {
```
leonsenft: Did you consider using a visitor to convert the AST, rather than manually performing `is` checks for each AST type? Any reason such an approach wouldn't work? I think it would make this a lot cleaner.

angular_analyzer_plugin/lib/src/converter.dart
```diff
@@ -397,78 +397,41 @@ class HtmlTreeConverter {
...
-      SourceRange closingSpan;
-      SourceRange closingNameSpan;
 
-      if (node.isSynthetic) {
-        openingSpan = _toSourceRange(closeComponent.beginToken.offset, 0);
-        openingNameSpan = openingSpan;
-      } else {
-        openingSpan = _toSourceRange(
-            node.beginToken.offset, node.endToken.end - node.beginToken.offset);
+      if (!node.isSynthetic) {
```
leonsenft: `node is ParsedElementAst` would effectively be the same check right? This would be cleaner too since you wouldn't need the `as ParsedElementAst` casts below due to type promotion.

angular_analyzer_plugin/lib/src/converter.dart
```diff
@@ -397,78 +397,41 @@ class HtmlTreeConverter {
...
-        openingSpan,
-        closingSpan,
-        openingNameSpan,
-        closingNameSpan,
         attributes,
-        findTemplateAttribute(attributes),
+        node.closeComplement,
         parent,
-        isTemplate: false,
+        openingNameSpanOverride: openingNameSpan,
```
leonsenft: Mind commenting what this is for and why it's only needed for `ElementAst`?

angular_analyzer_plugin/lib/src/converter.dart
```diff
@@ -903,6 +797,75 @@ class HtmlTreeConverter {
...
+    final isTemplate = tagName == 'template';
+    SourceRange openingSpan;
+    SourceRange openingNameSpan;
+    SourceRange closingSpan;
+    SourceRange closingNameSpan;
+
+    openingNameSpan = openingNameSpanOverride;
+
+    if (node.isSynthetic) {
+      openingSpan = _toSourceRange(closeComplement.beginToken.offset, 0);
```
leonsenft: This code to me looks like synthetic elements only occur when the opening tag is missing (since we read the close complement token). Is it never the case that a synthetic element occurs due to a missing closing tag?

angular_analyzer_plugin/lib/src/converter.dart
```diff
@@ -903,6 +797,75 @@ class HtmlTreeConverter {
...
+      openingNameSpan ??= openingSpan;
+    } else {
+      openingSpan = _toSourceRange(
+          node.beginToken.offset, node.endToken.end - node.beginToken.offset);
+      openingNameSpan ??=
+          new SourceRange(node.beginToken.offset + '<'.length, tagName.length);
+    }
+
+    if (closeComplement != null) {
```
leonsenft: Is this null check necessary? It's accessed unguarded on line 816. Is there some implicit guarantee that it's non-null when the node is synthetic?

angular_analyzer_plugin/lib/src/converter.dart
```diff
@@ -903,6 +797,75 @@ class HtmlTreeConverter {
...
+    );
+
+    for (final attribute in attributes) {
+      attribute.parent = element;
+    }
+
+    final children = _convertChildren(node, element);
+    element.childNodes.addAll(children);
+
+    if (!element.isSynthetic &&
```
leonsenft: Mind commenting (or using well-named local variables) to document what this check does? It's not very obvious but I think it adds an empty text child when there's nothing between the opening and closing tags? Why is this even necessary?

angular_analyzer_plugin/lib/src/resolver.dart
MichaelRFairhurst: They all are.

Until I can store my own resolution information inside of AngularAst, I am using my own AST. That didn't, and still doesn't, distinguish special tags like `<template>`, `<ng-content>`, etc.

I've definitely considered changing this, but I haven't done it yet.

angular_analyzer_plugin/lib/src/converter.dart
MichaelRFairhurst: Yep, this code used to iterate over the trees produced by package:html. It was adjusted to walk the angular_ast trees, but it hasn't been switched into a visitor.

I don't think it'd be particularly hard, so I could probably do it in a follow-up CL, (or even, a precursor CL), but I don't think I have enough cycles to make it worth it right now.

angular_analyzer_plugin/lib/src/converter.dart
MichaelRFairhurst: Great idea, will do!

angular_analyzer_plugin/lib/src/converter.dart
MichaelRFairhurst: I think correct, in that case the CloseComplement will be synthetic..?

angular_analyzer_plugin/lib/src/converter.dart
MichaelRFairhurst: Yeah. I could add an assert there!

angular_analyzer_plugin/lib/src/converter.dart
MichaelRFairhurst: Will do -- its for autocomplete. If this, if the user is targeting a text node we can give them suggestions, and that handles both `<div>^</div>` as well as `<div> ... ^ ... </div>`.

angular_analyzer_plugin/lib/src/resolver.dart
leonsenft: Sorry I think I still need further clarification. So any HTML tag (no `-`), plus any special Angular tags are considered as "HTML tags", but otherwise if it has a `-` it's considered something else, like a custom element?

angular_analyzer_plugin/lib/src/converter.dart
leonsenft: Ah that makes sense. Agreed, no need to update now. If you had time, doing it first would make this change easier to grep, but since you don't I'd leave it and just file a low priority issue to refactor. :+1:

angular_analyzer_plugin/lib/src/converter.dart
leonsenft: Oh right! This check is only checking if the opening tag is synthetic. Since the opening tag contains the closing tag I was conflating the meaning.

angular_analyzer_plugin/lib/src/converter.dart
MichaelRFairhurst: Thanks for challenging this!

These previously were in four places, with subtle differences between them, that I had to reason about & condense down. I thought this and ng-content needed their own logic.

But the case I _thought_ this covered (`< div`) is not actually a valid case. I just tried commenting out the code that _uses_ `openingNameSpanOverride`, and all tests passed (meaning both `ElementAst` and `EmbeddedContentAst` were doing things differently for no reason), so this can be removed entirely.

angular_analyzer_plugin/lib/src/converter.dart
MichaelRFairhurst: Rather than an assert, I'll just check closeComplement != null here too. That's the most defensive. If we ever get a null closeComplement, this branch cannot successfully run, but the next one can.

angular_analyzer_plugin/lib/src/resolver.dart
MichaelRFairhurst: Yep, exactly.

I plan to relax the '-' at some point in the future. I've had issues where dart:html didn't include standard references to every type of regular element. So I turned this off only to find things like `<audio>` were suddenly flagged. But I think I've caught all of them, and until I relax this, `<audioo` won't be flagged.

More here: #248

MichaelRFairhurst:
ok feedback finally addressed! PTAL @leonsenft

angular_analyzer_plugin/lib/src/converter.dart
```diff
@@ -828,10 +818,9 @@ class HtmlTreeConverter {
             closeComplement.endToken.end - closeComplement.beginToken.offset);
         closingNameSpan =
             new SourceRange(closingSpan.offset + '</'.length, tagName.length);
-      } else if (isTemplate) {
-        // Close range for <template /> tags
-        closingSpan = _toSourceRange(node.endToken.end, 0);
-        closingNameSpan = closingSpan;
+      } else {
+        // TODO(mfairhurst): generate a closingSpan for synthetic tags too. This
```
leonsenft: This sounds important, are you fine addressing in another CL?

MichaelRFairhurst: Yeah, I'm pretty unhappy with this so while all tests currently pass, I'm going to invest some into sharpening this system up shortly.

I'm thinking, assertions that the tree is "well-formed" for autocomplete. And then hopefully I can get the assertions passing under all cases, and then adapt auto-complete to work with the reliable structure it should then have...

PR #638, opened 2018-09-18
Fix #631 more disallowed operators

### LINKED ISSUE #631: warn using ++, += in a template
Looks like I need to vet which types of assignments, and which unary operators, are valid in dart templates and warn when they're used.

MichaelRFairhurst:


angular_analyzer_plugin/test/resolver_test.dart
```diff
@@ -283,40 +283,6 @@ class TestPanel {
     // no assertion. Just don't crash.
   }
 
-  // ignore: non_constant_identifier_names
```
leonsenft: Did you refactor/shuffle a large number of tests with this change? If so could I request you please do that as part of a separate CL so I can focus on which test cases apply directly to this change?

angular_analyzer_plugin/test/resolver_test.dart
```diff
@@ -5264,6 +5395,7 @@ class TestPanel {
         "return 5");
   }
 
+  // ignore: non_constant_identifier_names
```
leonsenft: nit: I noticed there's some funniness here, where this `ignore` comment is both on the return type and the actual identifier on the next line.

angular_analyzer_plugin/test/resolver_test.dart
MichaelRFairhurst: Yeah, I _thought_ doing both at once was a bad idea, and this diff really truly is just awful. Split that out into #639, which is landed, and rebased.

MichaelRFairhurst:
@leonsenft yeah, I suppose if I were a full time member of the angular team I probably would have put the effort into supporting compound assignment and unary operators.

That said, if we don't have support, I _definitely_ want to catch it.

For instance, `--x` is treated as `-(-x)`!! So no error for using it, and it becomes a noop! Very confusing.

angular_analyzer_plugin/test/resolver_test.dart
leonsenft: Thanks!

PR #643, opened 2018-09-27
Modify AngularErrorVerifier to pass InheritanceManager2 to analyzer.

stereotype441:
This is needed due to the change in the ErrorVerifier constructor in
https://github.com/dart-lang/sdk/commit/8a2cfd86139c58c3378d903d76c43663c1c61b29.

### DIFF
--- a/angular_analyzer_plugin/lib/src/resolver.dart
+++ b/angular_analyzer_plugin/lib/src/resolver.dart
@@ -1565,7 +1565,7 @@ class SingleScopeResolver extends AngularScopeVisitor {
         library,
         typeProvider,
         new InheritanceManager(library),
-        new InheritanceManager2(typeSystem),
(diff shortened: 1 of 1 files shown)


MichaelRFairhurst:
looks like a missing cast (since I have implicit downcasts turned off) and some test code we missed..

stereotype441:
@MichaelRFairhurst Can you recommend a fix?  I see how to fix the missing cast but I don't understand the errors in the test code

MichaelRFairhurst:
Oh, I misread the test failures.

Will  fix those in a separate CL, that's an unrelated breakage.

stereotype441:
Ok, I believe I've addressed the error that's due to my CL.  Should I go ahead and land this, or wait for your fix?

MichaelRFairhurst:
I landed #644, retesting, but I'm pretty positive this will work.

PR #653, opened 2018-10-29
Change behavior to not override analyzer classes where not necessary.

MichaelRFairhurst:
Work required on the analyzer side to remove `AngularErrorVerifier` and
`AngularResolver`, but this now decouples most of the rest of the behavior
so that removing the former should be simple. Harder will be removing
pipes so that `AngularResolver` can be removed, but that's in the works
too.


MichaelRFairhurst:
blocked on https://github.com/dart-lang/angular_analyzer_plugin/pull/654 it seems

PR #660, opened 2018-11-03
Use `LocalElementBuilder` so that `FunctionExpression`s analyze.

MichaelRFairhurst:
Then we don't have to override `ErrorVerifier`, reducing coupling.


MichaelRFairhurst:
Looks like travis is failing because it doesn't test against the analyzer branch.

I'll see if I can fix that first.

PR #664, opened 2018-11-15
Fix issues in not-exported errors, more tests.

MichaelRFairhurst:
Previous non-error tests didn't actually assert no errors were reported,
only that the elements were resolved. That worked under the previous
approach where the library scope was restricted, but that doesn't work
under the new aproach where resolution occurs regardless of the export
error.

After enabling the export error, the biggest issues found were:
* summarized ExportedIdentifier has '' for prefix, not null, when
missing
* identifer.name on a PrefixedIdentifier yields prefix.suffix rather
than just suffix.

One other edge case bug remained (`prefixed.topLevelFn()`), and then
added some tests for other edge cases, and did some other minor
refactoring like introducing new locals for better type promotion.

### DIFF
--- a/angular_analyzer_plugin/lib/src/resolver.dart
+++ b/angular_analyzer_plugin/lib/src/resolver.dart
@@ -266,13 +266,13 @@ class AngularSubsetVisitor extends RecursiveAstVisitor<Object> {
   /// Flag the rest and give the hint that they should be exported.
   void visitIdentifier(Identifier id) {
     final element = id.staticElement;
+    final parent = id.parent;
     if (id is PrefixedIdentifier && id.prefix.staticElement is! PrefixElement) {
       // Static methods, enums, etc. Check the LHS.
       visitIdentifier(id.prefix);
       return;
     }
-    if (id.parent is PropertyAccess &&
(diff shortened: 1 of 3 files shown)


angular_analyzer_plugin/lib/src/resolver.dart
```diff
@@ -266,13 +266,13 @@ class AngularSubsetVisitor extends RecursiveAstVisitor<Object> {
...
     final element = id.staticElement;
+    final parent = id.parent;
     if (id is PrefixedIdentifier && id.prefix.staticElement is! PrefixElement) {
       // Static methods, enums, etc. Check the LHS.
       visitIdentifier(id.prefix);
       return;
     }
-    if (id.parent is PropertyAccess &&
-        identical(id, (id.parent as PropertyAccess).propertyName)) {
+    if (parent is PropertyAccess && identical(id, parent.propertyName)) {
```
scheglov: `AstNode` does not implement `==`, so it does the same as `identical`.

MichaelRFairhurst: Thanks!

PR #665, opened 2018-11-15
Fix #641, named arguments are now allowed in angular

### LINKED ISSUE #641: Remove warning for named arguments
Reproduction:
```dart
@Component(
  selector: 'app',
  template: '''
    {{quote(value, character: "'")}}
                   ^^^^^^^^^^^^^^
                   Named arguments not allowed in angular templates [disallowed_expression]
  '''
)
class AppComponent {
  final = "Hello world!";

  String quote(String value, {this.character: '"'}) => value;
}
```

Support for named arguments was added in version 5.0.0.

MichaelRFairhurst:


MichaelRFairhurst:
Yep! The analyzer handles them by default, so it's just an additional check to remove. I had simply forgotten to do it for a long time..!

And yep, I didn't make any changes that would stop named arguments, so technically they've been suggested this whole time, would just get flagged if you ever used them.

PR #673, opened 2018-12-06
Remove an unnecessary and problematic cast

bwilkerson:
The cast refers to a private class that is being renamed. Removing the cast doesn't have any effect because the invoked method only requires a `TypeProvider`.


bwilkerson:
Yes. I'm trying to rename that class and running into conflicts. I'd be interested to understand why there are multiple branches and how to work with them.

PR #678, opened 2019-01-18
Update to the latest version of analyzer and replace uses of deprecated APIs

bwilkerson:


angular_analyzer_plugin/analysis_options.yaml
```diff
@@ -8,7 +8,6 @@ analyzer:
 
 linter:
   rules:
-    - prefer_final_locals
```
MichaelRFairhurst: Hmm, I really do prefer final locals, was there a reason to remove this?

angular_analyzer_plugin/lib/src/resolver.dart
```diff
@@ -1645,7 +1645,8 @@ class SingleScopeResolver extends AngularScopeVisitor {
     // half-complete-code case: ensure the expression is actually there
     if (attribute.expression != null &&
         !typeSystem.isAssignableTo(
-            attribute.expression.bestType, typeProvider.boolType)) {
+            attribute.expression.staticType ?? typeProvider.dynamicType,
```
MichaelRFairhurst: I think the case where staticType is null has to be part of the condition before isAssignableTo. Otherwise, if implicit downcasts are turned off, the assignability check will fail and they'll get a spurious secondary error.

If this is how `bestType` already worked though, I'm OK with doing that myself in a followup CL.

angular_analyzer_plugin/lib/src/resolver.dart
```diff
@@ -1787,7 +1788,8 @@ class SingleScopeResolver extends AngularScopeVisitor {
       // half-complete-code case: ensure the expression is actually there
       if (attribute.expression != null &&
           !typeSystem.isAssignableTo(
-              attribute.expression.bestType, typeProvider.numType)) {
+              attribute.expression.staticType ?? typeProvider.dynamicType,
```
MichaelRFairhurst: Same as above

angular_analyzer_plugin/lib/src/resolver.dart
```diff
@@ -1830,13 +1832,17 @@ class SingleScopeResolver extends AngularScopeVisitor {
           // half-complete-code case: ensure the expression is actually there
           if (attribute.expression != null &&
               !typeSystem.isAssignableTo(
-                  eventType, attribute.expression.bestType)) {
+                  eventType,
+                  attribute.expression.staticType ??
+                      typeProvider.dynamicType)) {
```
MichaelRFairhurst: Same as above

angular_analyzer_plugin/lib/src/resolver.dart
```diff
@@ -1830,13 +1832,17 @@ class SingleScopeResolver extends AngularScopeVisitor {
...
             errorListener.onError(new AnalysisError(
                 templateSource,
                 attribute.valueOffset,
                 attribute.value.length,
-                AngularWarningCode.TWO_WAY_BINDING_OUTPUT_TYPE_ERROR,
-                [output.eventType, attribute.expression.bestType]));
+                AngularWarningCode.TWO_WAY_BINDING_OUTPUT_TYPE_ERROR, [
+              output.eventType,
+              attribute.expression.staticType ?? typeProvider.dynamicType
```
MichaelRFairhurst: this could then be removed

angular_analyzer_plugin/lib/src/resolver.dart
```diff
@@ -1858,7 +1864,7 @@ class SingleScopeResolver extends AngularScopeVisitor {
       ExpressionBoundAttribute attr, InputElement input) {
     // half-complete-code case: ensure the expression is actually there
     if (attr.expression != null) {
-      final attrType = attr.expression.bestType;
+      final attrType = attr.expression.staticType ?? typeProvider.dynamicType;
```
MichaelRFairhurst: Same as above but more indirectly

angular_analyzer_plugin/analysis_options.yaml
bwilkerson: The same rule was in the list in two places. The second was sorted so I chose to remove this one. But the lint is still enabled.

angular_analyzer_plugin/lib/src/resolver.dart
bwilkerson: Yes, `bestType => staticType ?? typeProvider.dynamicType`

PR #681, opened 2019-02-04
Summarize syntactic model

MichaelRFairhurst:
The last commit of this adds a way to summarize the syntactic submodel defined in #680.


MichaelRFairhurst:
Easiest way to view the diff: https://github.com/dart-lang/angular_analyzer_plugin/pull/681/commits/ebc5798e2b7362178597c6530b63ac9ce252a88e

This raises two questions re #680:
- do I need a syntactic submodel, or should I simply build these classes when scanning dart sources?
- do I need `ReferenceOrList`?

PR #688, opened 2019-02-28
Remove importUri from test.

stereotype441:
The corresponding getter was removed from analyzer_plugin in
https://github.com/dart-lang/sdk/commit/5e2529c82e11150dbc2448f0f43aa4fcea16177b

### DIFF
--- a/angular_analyzer_plugin/test/completion_contributor_test_util.dart
+++ b/angular_analyzer_plugin/test/completion_contributor_test_util.dart
@@ -167,7 +167,6 @@ abstract class BaseCompletionContributorTest extends AbstractAngularTest {
     } else {
       expect(cs.relevance, equals(relevance), reason: completion);
     }
-    expect(cs.importUri, importUri);
     expect(cs.selectionOffset, equals(completion.length));
     expect(cs.selectionLength, equals(0));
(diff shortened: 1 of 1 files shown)


angular_analyzer_plugin/test/completion_contributor_test_util.dart
```diff
@@ -167,7 +167,6 @@ abstract class BaseCompletionContributorTest extends AbstractAngularTest {
     } else {
       expect(cs.relevance, equals(relevance), reason: completion);
     }
-    expect(cs.importUri, importUri);
```
stereotype441: @scheglov @MichaelRFairhurst is there a new way I should be testing this functionality?

MichaelRFairhurst: It looks like this parameter isn't specified/used anywhere. I think this was probably forked from analysis server test code, and that probably needed to support this at one time (and has presumably been migrated). I think for the plugin removing it is all we need.

I want to remove the parameters too since they do nothing, but that is secondary to merging this PR.

MichaelRFairhurst:
Thanks Paul/Konstantin!

PR #690, opened 2019-03-07
Use SetOrMapLiteral instead of MapLiteral

MichaelRFairhurst:


MichaelRFairhurst:
closed in favor of #691  (I made this one with the wrong branch name) which is now merged.

Issue #687, opened 2019-02-24
What's about to export AngularDriver class to be used outside?


guid-empty:
Hi!

Is it possible to export AngularDriver class from this packages to be used as one of available classes?

One of the cases is ability to check custom angular template in build pipeline, for example:

`
      final plugin = AngularAnalyzerPlugin(PhysicalResourceProvider.INSTANCE);

      final versionCheckParams =
          new PluginVersionCheckParams('~/.dartServer/.analysis-driver', '/usr/lib/dart', '2.1.1');
      await plugin.handlePluginVersionCheck(versionCheckParams);

      final root = ContextRoot(rootPath, [], optionsFile: optionsPath);

      plugin.start(PluginCommunicationChannelStub());
      final genericDriver = plugin.createAnalysisDriver(root);
      final angularDriver = genericDriver as AngularDriver;

      genericDriver.addFile(angularTemplatePath);
      //  await genericDriver.performWork();

      final htmlResult = await angularDriver.requestHtmlResult(angularTemplatePath);

      print('angular template has errors -> ${htmlResult.errors.isNotEmpty}, ${dartResult.errors.isNotEmpty}');

      if (htmlResult.errors.isNotEmpty) {
        htmlResult.errors.forEach((AnalysisError error) {
          print(error.message);
          print(error.correction);
        });
      }
`

Thanks.

PR #692, opened 2019-03-11
export angular_driver to be used outside as external tool

guid-empty:
Hello!

This pull request is intended to close this issue: https://github.com/dart-lang/angular_analyzer_plugin/issues/687.

AngularDriver class (requestHtmlResult, requestDartResult methods) can be used outside the plugin to check quality of angular templates as external tool, for example, as hook to not commit dirty and not approved code to repository. Or to check the code quality in CI process before its publishing to production.

Please review and approve this export.

Regards,
Andrey Smirnov


MichaelRFairhurst:
Thanks for the PR!

Really cool to see you are using it for this!

The API here is not stable, and in fact, is being refactored pretty heavily right now: https://github.com/dart-lang/angular_analyzer_plugin/compare/master...MichaelRFairhurst:for-google3#diff-3c5761f3b543b754d8b895d9c12c9198

I would say for now I would encourage importing an exact version and then use from src. (you have my permission because I know I've warned you it may change soon!)

After that refactor, I can stabilize the API for you! Two options there:
* I can accept this PR but not publish until the refactor is done
* I can reach out to you to reexamine this PR once the refactor is done

Do you have a preference there, and does that all sound good to you?

Like I said it's really cool to see you are planning on doing this! Let me know how I can make sure you get what you need from the code that will exist now and going forward :)

kevmoo:
@guid-empty – all development for the plugin has moved to https://github.com/dart-lang/angular/tree/master/angular_analyzer_plugin

Please open a discussion there on the best way forward!

PR #697, opened 2019-04-04
Don't report "view range" in HTML navigation

MichaelRFairhurst:


angular_analyzer_plugin/pubspec.yaml
```diff
@@ -9,11 +9,11 @@ homepage: https://github.com/dart-lang/angular_analyzer_plugin
 environment:
   sdk: '>=2.0.0-dev.0.0 <3.0.0'
 dependencies:
-  analyzer: '0.34.2'
+  analyzer: '0.34.3'
   plugin: '^0.2.0'
   #  tuple: '^1.0.1' Does not yet support Dart 2
   analyzer_plugin: '0.0.1-alpha.6'
-  angular_ast: '^0.5.0'
+  angular_ast: any
```
bwilkerson: That's rather brave. Are you sure the pieces you need won't break with later versions?

MichaelRFairhurst: Oops! This was fooling around earlier that got committed....reverted!!

PR #702, opened 2019-04-05
Roll analyzer

MichaelRFairhurst:
This mostly just merges SDK_AT_HEAD into master, which contains the updates to the plugin to work with the latest analyzer code.

Otherwise, it updates the pubspec to import the latest.

This was blocked on publishing a new version of angular_ast, which was blocked on some angular infrastructure.


MichaelRFairhurst:
ping @bwilkerson

angular_analyzer_plugin/lib/src/resolver.dart
```diff
@@ -350,23 +350,23 @@ class AngularSubsetVisitor extends RecursiveAstVisitor<Object> {
       // Don't visit the TypeName or it may suggest exporting it, which is not
       // possible.e.
 
-      list.elements.accept(this);
+      list.elements2.accept(this);
```
bwilkerson: Can we jump ahead to the point where `elements2` becomes `elements` again?

PR #707, opened 2019-04-30
Remove FromFilePrefixError class

MichaelRFairhurst:


MichaelRFairhurst:
Yep! I had assumed I'd had a better reason for creating this class originally. Looks like I probably just didn't see the `forValues` constructor originally.

